PR #31, opened 2019-11-19
Continous api

pronapro:
- What does this PR do?
creates API endpoint for the forecasting model

- What Jiira task is related to this PR?

- How do I test out this PR?
git app.py
- Which endpoints should be ready for testing?
local URL:/prediction
issue: it is so slow to load
Heroku Url:https://firstapp-napro.herokuapp.com/
issue: fails with a Request timeout error, I have to find a way to reduce the time the model takes to predict


Baalmart:
> Wow, these are too many files changed. It is a good practice to keep it below 20 files changed unless you are just setting up a project.
> 
> Also, could this be a stale PR? Do we need to delete it?

@pronapro , could you be having any update about this? We are trying to reduce the number of active PRs/branches on our central remote repository.

sserurich:
> @sserurich should we close this PR? Is it still valid?
> Too many changed files : 5000 of them!
> cc: @pronapro

@Baalmart , we can archive this PR.. @pronapro , hope you're in agreement with me.

PR #39, opened 2020-01-14
Data mgt and incentives

Baalmart:
- What does this PR do?
This PR created a new endpoint for sending mobile money to the device host's phone number.
The PR also shares the app structure for the data management service with a main focus on the weather forecast data as the third party data going to be stored and accessed via an API.
The link to the JIRA board task is [here](https://airqoteam.atlassian.net/browse/APA-8).
- How do I test out this PR?
Git clone this repo `cd AirQo-api/src/incetives-service` Then run `npm install` to get all the necessary dependencies Afterwards run `npm run test` to test out the end points using any REST client tool available to you.
partial url: /api/v1/incentives/momo. More details about how to run this endpoint can be in [AirQo API documentation](https://docs.google.com/spreadsheets/d/1-NGTDBtTcIi8Fg2UYSWjcK6geKrZdopz5kxRFsCEMPs/edit#gid=530501335)


src/incetives-service/app.yaml
```diff
@@ -0,0 +1,6 @@
+runtime: nodejs10
+service: incentives
+env_variables:
```
Baalmart: These keys should not be visible in the codebase. Need to utilize good security practices here.

src/device-registry/.env
```diff
@@ -1,3 +1,5 @@
 PROJECT_ID=airqo-250220
```
Baalmart: This file should not be part of the commits made. Utilize .gitignore file for this or investigate in case it is already there.

PR #50, opened 2020-03-18
adding k8s deployment scripts for four microservices to begin with 

danielogen:
**What does this PR do?**
It adds k8s deployment scripts for the different microservices
N/A
A test can run on Jenkins server by triggering a CI/CD pipeline from GitHub
No endpoints available but the test can be done on Jenkins server


danielogen:
I need to take a closure look at the file an make a new PR.

PR #51, opened 2020-03-19
GKE deployment script - four microservices to begin with

danielogen:
**What does this PR do?**
It adds k8s deployment scripts for the different microservices
N/A
A test can run on Jenkins server by triggering a CI/CD pipeline from GitHub
No endpoints available but the test can be done on Jenkins server


k8s/airqo-deployments.yaml
```diff
@@ -0,0 +1,200 @@
...
+    app: airqo-analytics-backend
+  ports:
+    - protocol: TCP
+      port: 5000
+      targetPort: 5000
+      nodePort: 30004
+  type: NodePort
+
+####################################################################
+# Start analytics frontend microserive deployment and provisioning
```
Baalmart: Thanks Daniel, good to see the PORTS configurations for the different services. Awesome stuff. 
I am just wondering about the inclusion of "analytics frontend" config details in this backend repo, could you shed some light on this?

danielogen: Ah ok, I was writing the scripts under the impression that all the K8S are supposed to be in a single file. I forgot to consider the repo differences :). Kindly advice! Should we create a "k8s" folder in the "AirQ0-frontend" repo?

Baalmart: OoH, I get your point.

I am thinking that if we are to consider different build pipelines for both backend and frontend, we might need to created a k8s file for "AirQo-frontend". My imagination is that the tests for frontend are different from those of the frontend. Example: UI Selenium tests for the frontend are not required for the backend.....

Could you look into this further and share your new thoughts? Thanks again bro! :)

danielogen: Absolutely right, I think it will be appropriate to create a "k8s" folder in "AirQo-frontend" repo, which will contain the deployment scripts specific to frontend services. Nice catch :).

k8s/airqo-deployments.yaml
```diff
@@ -0,0 +1,160 @@
...
+    - protocol: TCP
+      port: 5000
+      targetPort: 5000
+      nodePort: 30003
+  type: NodePort
+###################################################################
+# Start analytics backend microserive deployment and provisioning
+###################################################################
+---
+apiVersion: apps/v1beta1
```
sserurich: wondering why apiVersion on some microservices deployment is app/v1beta1 and others it just app/v1, also  consistency between deployments and services for the apiVersion.

danielogen: I will probably have to change the deployment versions to be uniform but for now, any (`apps/v1` or `apps/v1beta1`) will work. The `services` format is always different from the `deployments`, hence the inconsistency.

k8s/airqo-deployments.yaml
```diff
@@ -0,0 +1,160 @@
...
+        - name: airqo-datamgt-service
+          image: us.gcr.io/airqo-250220/airqo-datamgt-service:latest
+          imagePullPolicy: Always
+          ports:
+            - containerPort: 3000
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: airqo-datamgt-service-svc
```
Baalmart, line 71: A renaming suggestion: **airqo-data-mgt-service**. 

How does this sound?

sserurich:
> Who is doing the merge duty for this one? Should I go ahead and do it?@martin, you can go ahead, though I have noticed there some comments that need to be resolved.

k8s/airqo-deployments.yaml
danielogen: Sounds good. Comment implemented

Baalmart:
> > Who is doing the merge duty for this one? Should I go ahead and do it?@martin, you can go ahead, though I have noticed there some comments that need to be resolved.

Just seen this. Could we first make sure that it has the latest changes from MASTER?

danielogen:
Hi @Baalmart, Looks like this PR contain only the requested changes from the reviewers. The rest of the content is available [here](https://github.com/airqo-platform/AirQo-api/blob/master/k8s/airqo-deployments.yaml). I will update the PR to include `locate`, `locate-registry` and `device-monitoring` micro-services. Thanks!

Baalmart:
> Hi @Baalmart, Looks like this PR contain only the requested changes from the reviewers. The rest of the content is available [here](https://github.com/airqo-platform/AirQo-api/blob/master/k8s/airqo-deployments.yaml). I will update the PR to include `locate`, `locate-registry` and `device-monitoring` micro-services. Thanks!

Cool, thanks.

danielogen:
Check out the updated PR here https://github.com/airqo-platform/AirQo-api/pull/105

PR #52, opened 2020-03-24
Retrieving & saving data from clarity api to mongodb and endpoints for retrieving data from mongodb

sserurich:


sserurich:
> 1. Can we have comments in every method regarding what it does?
> 2. Is this a local mongoDB being used?
> 3. If its publicly available, is it already populated?
> 4. The clarity api's default limit is 500. How is this being handled?

1. am going to work on the comments and will be updating it
2. I used mongodb and am currently exploring exporting it to cloud or shared db (@daniel will be setting up instance)
3. the db is already populated upto yesterday's measurements (hourly at around 10 or 9pm and daily upto 22nd march) hope to set up service that continously updates these after having a shared db.
4. The clarity api's default is 500 but you can set limit to 20000 and this way I was able to get all the hourly and daily measurements since 2019-09-27T13:00:00Z when devices started monitoring.

src/analytics/controllers/controller.py
```diff
@@ -1,11 +1,70 @@
...
+from helpers.clarity_api import ClarityApi
+from bson import json_util, ObjectId
+import json
 
 _logger = logging.getLogger(__name__)
 
 analytics_app = Blueprint('analytics_app', __name__)
 
+
+@analytics_app.route('/api/v1/device/measurements', methods=['GET'])
```
danielogen: Is it possible to move all the clarity partial URLs to one python file say for example "constant.py" and just import it in the controller.py? It will make the routes look prettier :)

src/analytics/config.py
```diff
@@ -11,19 +11,28 @@ class Config:
     SECRET_KEY = 'sample-code-local-environment'
     #SERVER_PORT = 5000
 
+    CLARITY_API_BASE_URL= "https://clarity-data-api.clarity.io/v1/"
```
Baalmart: We need to be aware of the security aspects of our system. We could explore the usage of environmental variables.

PR #55, opened 2020-03-26
setting up Flask for AI/ML model utilization

Baalmart:
- What does this PR do?
This PR updates the ml-service directory so that it can be used to consume ML models and etc. 
There is no JIRA card for this task.
- How do I test out this PR?
Git clone this repo `git pull` and then `cd AirQo-api/src/ml-service` Then run all the necessary set up tasks for Flask apps.
There are no endpoints for this PR.
- Quick notes about our Git workflow?
Please refer to [these notes](https://github.com/airqo-platform/AirQo-api/wiki). Special attention should be given to the part of how we create branches and the need to always work outside the **master** branch. And to always ensure that we have the latest master branch code within each branch we are operating it.


Baalmart:
Since this is just foundational/structure code, I am just about to merge in the interest of time. :)

sserurich:
it looks good, the only small issue  is that the name module (folder named modules) refers to file name in python, and it might cause some confusion

PR #60, opened 2020-04-29
Analytics Backend Service

sserurich:
Service can be tested out by fetching the branch locally and following the instructions in README.txt


src/analytics/models/dashboard.py
```diff
@@ -0,0 +1,46 @@
+import app 
+from helpers import mongo_helpers
```
Baalmart, line 2: Ideally, this is where we are supposed to define all our collection schemas. As seen in ATLAS, we have: 

1. device_daily_historical_averages
2. device_daily_measurements
3. device_hourly_measurements
4. device_raw_measurements
5. devices
6. division
7. monitoring_site

So those are the tables (collections) whose schemas are supposed to be under this `models` directory. 
More like:
```
from .db import db

class Movie(db.Document):
Division = db.StringField(required=True, unique=True)
SiteName = db.StringField(required=True, unique=True)
DeviceCode = db.StringField(required=True, unique=True)
Latitude = db.StringField(required=True, unique=True)
Longitude = db.StringField(required=True, unique=True)
Organisation = db.StringField(required=True, unique=True)
LocationCode = db.StringField(required=True, unique=True)
Parish = db.StringField(required=True, unique=True)
LatestHourlyMeasurement = db.ListField(db.StringField(), required=True)
```

Baalmart: And then within the same class, we could add some "static" or "instance" functions which are currently under `helpers/mongo_helpers`

src/analytics/models/graph.py
```diff
@@ -0,0 +1,103 @@
+import app 
+from helpers import mongo_helpers, helpers
```
Baalmart, line 2: Could we define the schema here and have most of this business logic go under `controllers/Graph.py` ?

src/analytics/models/monitoring_site.py
```diff
@@ -0,0 +1,198 @@
+import app
```
Baalmart, line 1: Could we also define the schemas from here? And move most of the business logic to `controllers` directory?

src/ml-service/packages/ml-api/met-forecast.py
```diff
@@ -0,0 +1,226 @@
+import base64
```
Baalmart, line 1: This could potentially go into the data management service. Right?

src/ml-service/packages/ml-api/met-forecast.py
```diff
@@ -0,0 +1,226 @@
...
+    rows_to_insert = weather_forecasts
+    errors = client.insert_rows(table, rows_to_insert)
+    if errors == []:
+        return 'Records saved successfully.'
+    else:
+        return errors
+    
+  
+def get_location_hourly_weather_forecasts(latitude:float, longitude:float):
+        headers = { 'x-ibm-client-id': MET_API_CLIENT_ID,
```
Baalmart, line 51: We need to document these endpoints in our API documentation.

src/ml-service/packages/ml-api/met-forecast.py
```diff
@@ -0,0 +1,226 @@
...
+                    location = features_properties['location'].get('name')
+                    location_lat = features_geometry.get('coordinates')[1]    
+                    location_long = features_geometry.get('coordinates')[0]
+                    point_elevation = features_geometry.get('coordinates')[2]
+                    distance_from_requested_point = features_properties['requestPointDistance']
+                    created_at  =  datetime.now()
+
+                    if forecast_timeseries:
+                        for i in range(0, len(forecast_timeseries)):
+                            location_name = location
```
Baalmart, line 101: This looks a little bit wordy, could we be able to use `destructuring` when assigning these variables?

src/ml-service/packages/ml-api/met-forecast.py
```diff
@@ -0,0 +1,226 @@
...
+                forecast_timeseries = features_properties['timeSeries']
+                location = features_properties['location'].get('name')
+                location_lat = features_geometry.get('coordinates')[1]    
+                location_long = features_geometry.get('coordinates')[0]
+                point_elevation = features_geometry.get('coordinates')[2]
+                distance_from_requested_point = features_properties['requestPointDistance']
+                created_at  =  datetime.now()
+
+                if forecast_timeseries:
+                    for i in range(0, len(forecast_timeseries)):
```
Baalmart, line 173: same here, we could explore `destructuring `when assigning values to variables from the same "source:"

Baalmart:
> * We need to be able to utilize different instances of DBMS when testing in different development environments
> * There is too much code duplication, we need to exercise OOP or SOLID principles accordingly. There were too many endpoints for saving a report.

Also, carry out error handling on all our endpoints accordingly.

PR #128, opened 2020-08-06
Api device maintenance

Priscilla-AL:
What does this PR do?
adds isDueMaintainace field to out put
What JIRA task is related to this PR?
https://airqoteam.atlassian.net/jira/software/projects/NET/boards/28?assignee=5d91eed256bb950dd6489270&selectedIssue=NET-5
How do I test out this PR?
Git clone this repo cd AirQo-api/src/device-monitoring  e.t.c

Endpoint: [GET] /api/v1/monitor/devices/online_offline

### DIFF
--- a/src/device-monitoring/controllers/device_status.py
+++ b/src/device-monitoring/controllers/device_status.py
@@ -245,18 +245,21 @@ def get_all_online_offline():
             devices_with_coordinates =[]
             for device in result['online_devices']:
                 if (device['latitude'] is not None) or (device['longitude'] is not None) :
-                    
-                    mapped_device = {
-                        'channelId':device['channelID'],
-                        'latitude': device['latitude'],
-                        'locationId': device['location_id'],
-                        'longitude': device['longitude'],
-                         'power': device['power'],
-                         'productName': device['product_name'],
-                         'phoneNumber': device['phoneNumber'],
-                         'isOnline': True
-                    }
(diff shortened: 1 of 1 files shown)


Priscilla-AL:
Requested changes made

lilly-muyama:
The isDueMaintenance field is not displaying accurate results. Some devices that are being shown as codeGreen should be codeRed. I tested 3 and they were all inaccurate

Priscilla-AL:
Yes, @lilly-muyama  I am aware. We agreed I would make further changes after you clarify with PO and users. for now its over due and not. Two categories.

lilly-muyama:
> 
> 
> Yes, @lilly-muyama I am aware. We agreed I would make further changes after you clarify with PO and users. for now its over due and not. Two categories.

@Priscilla-AL , my point is most of your devices are green which is not true. Most devices are overdue for maintenance. We can't have a map showing a lot of green nodes yet the top cards are showing the opposite. The endpoint is giving inaccurate results. The clarifications from the PO/users concern the frontend not the endpoint.

Priscilla-AL:
Yes @lilly-muyama, I am aware, as I mentioned its work in progress, but depending on your feedback I may need to change the back end as well or do the logic from the frontend. So the back end may be affected too. 
There is still a lot to be improved, Just thought you show demo the ring whatevers to account for my time.

PR #138, opened 2020-08-26
Updates cronjob for checking device status

sserurich:
**What does this PR do?**
  - This PR addresses the issues that have been preventing visuals i.e graphs & devices on netmanager for offline and online devices to display. It updates job for checking device status to use updated response from data-manager get last entry api.
   N/A
-  simply check the netmanager device-management visuals and map are showing correctly , since I updated job in the cloud 
or
 - git fetch branch fix-netmanager-device-status
 - change directory to device-monitoring (device monitoring service)
 - run pip install -r requirements.txt to get all the relevant packages.
 - Afterwards update mongo uri  in  jobs\check_device_status.py
-  then on the terminal (in the active environment) run python jobs\check_device_status.py and you should observe computations on the console.
- Confirm that when device management page is opened, map shows devices that are offline/online and piechart for offline/online devices shows up and the table listing offline devices.
   - n/a
image

### DIFF
--- a/src/device-monitoring/jobs/check_device_status.py
+++ b/src/device-monitoring/jobs/check_device_status.py
@@ -125,7 +125,7 @@ def get_device_channel_status():
 
         print(device_status_results)
 
-        #save_hourly_device_status_check_results(device_status_results)       
+        save_hourly_device_status_check_results(device_status_results)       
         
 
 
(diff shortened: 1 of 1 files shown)


lilly-muyama (approved):

Baalmart (commented):
Thanks @sserurich :

- [ ] After running the job as specified in the PR description, I get the following response:

image


Is that 👆 the expected response? If yes, please help me interpret it?

- [ ] As for the observations on the device management page, could you specify the environment under which these observations should be made? I assume production?

Baalmart (approved):
LGTM, though I failed to notice any computations in my console after running the cloud function.

PR #140, opened 2020-08-28
[Feature] Adding Device measurements

Baalmart:
- **What does PR do?**
This PR creates an endpoint for adding device (sensor) measurements to the platform.
Updates the Events collection and the corresponding API endpoint for adding time series data (measurements) of a device.

- **How do I test out this PR?**

Remember to use the _staging environment_ since it has the data you need to carry out your tests....
> For MAC: 

```
npm install
npm run stage-mac
```

> For PC:

```
npm install
npm run stage-pc
```

> Alternatively:

You can use Docker to test out.

- **What endpoints are ready for testing?**
Base URL:  http://localhost:3000/
Partial URL[POST]:  api/v1/devices/components/add/values
More details are in the [API documentation](https://docs.google.com/spreadsheets/d/1-NGTDBtTcIi8Fg2UYSWjcK6geKrZdopz5kxRFsCEMPs/edit#gid=897606332)


sserurich (changes requested):
Endpoint seems to work perfectly however, not sure how the values added to which device they belong because it's not specified anywhere in the request data. not sure whether it randomly assigns which device this data belongs to.
image

lilly-muyama (commented):
Works as shown in the api documentation. However, I need more explanation on the inputs and outputs of this endpoint and its use before I can approve.

danielogen (approved):
The endpoint works perfectly as described in the API doc.

Priscilla-AL (approved):
Works as expected.

Baalmart:
> Works as shown in the api documentation. However, I need more explanation on the inputs and outputs of this endpoint and its use before I can approve.

This endpoint is for adding sensor readings (can be PM, Humidity, Temperature, etc) for a component of a device.
So the inputs are:
- device name (as a query in the URL)
- component name (in the request body)
- readings of the component/sensor  (in the request body)
- timestamp of reading  (in the request body)

See REST client screenshot for more details:
image

Baalmart:
> Endpoint seems to work perfectly however, not sure how the values added to which device they belong because it's not specified anywhere in the request data. not sure whether it randomly assigns which device this data belongs to.
> image

Good observation!
Just updated the documentation - the partial URL requires one to add the device name as a query item - see screenshot below.
I also think that we need to cover more edge cases for this so that the device name is first looked up before proceeding with the operations, let me review asap....

Baalmart:
> Endpoint seems to work perfectly however, not sure how the values added to which device they belong because it's not specified anywhere in the request data. not sure whether it randomly assigns which device this data belongs to.
> image

This has been addressed, thanks for the review. Please checkout the 2 screenshots below:

**======Screenshot 1 (input validation)==========**

image

**======Screenshot 2 (response with correct input):=======**
image

Baalmart:
> > Endpoint seems to work perfectly however, not sure how the values added to which device they belong because it's not specified anywhere in the request data. not sure whether it randomly assigns which device this data belongs to.
> > image
> 
> This has been addressed, thanks for the review. Please checkout the 2 screenshots below:
> 
> **======Screenshot 1 (input validation)==========**
> 
> image
> 
> **======Screenshot 2 (response with correct input):=======**
> image

Hi @sserurich :
Just remaining with your review for us to finalise with this PR...

sserurich (approved):
the concern was addressed and PR works fine.

sserurich (changes requested):
after thorough examination of this api, here are some few things that might need to be updated: the component name, we need to know which component the values are being added to.. currently you only specify the component name which doesn't explicitly tell which component values belong to.. and also if same device name is called more than once, the old data is overridden. If you try to specify the component name in the request body, it doesn't come up in the response.
image

Baalmart:
> after thorough examination of this api, here are some few things that might need to be updated: the component name, we need to know which component the values are being added to.. currently you only specify the component name which doesn't explicitly tell which component values belong to.. and also if same device name is called more than once, the old data is overridden. If you try to specify the component name in the request body, it doesn't come up in the response.
> image
> 
> Also we have a challenge, the components we're adding values to as you see in image below, the quantities being measured are not specified while adding values for the component e.g the pms5003 component has two quantities for pm2.5 & pm10, those are not reflected in the api for adding values, mapping readings from thingspeak to this becomes challenging.
> image

Cool, thanks for the review. Let me address this asap...

Baalmart:
This has been addressed. To test, please test using the staging environment since it already has the device data. Let me also update the API documentation shortly.
Please check out the screenshots below:
image
image

=====================================
======================================

> after thorough examination of this api, here are some few things that might need to be updated: the component name, we need to know which component the values are being added to.. currently you only specify the component name which doesn't explicitly tell which component values belong to.. and also if same device name is called more than once, the old data is overridden. If you try to specify the component name in the request body, it doesn't come up in the response.
> image
> 
> Also we have a challenge, the components we're adding values to as you see in image below, the quantities being measured are not specified while adding values for the component e.g the pms5003 component has two quantities for pm2.5 & pm10, those are not reflected in the api for adding values, mapping readings from thingspeak to this becomes challenging.
> image

sserurich (approved):
looks good , concerns so far have been addressed.

PR #148, opened 2020-09-18
Ft location registry tenancy

lilly-muyama:
- What does this PR do?
This PR adds multi-tenancy to the location-registry microservice.
The link to the JIRA board task is [here](https://airqoteam.atlassian.net/jira/software/projects/PLAT/boards/41?selectedIssue=PLAT-73).
- How do I test out this PR?
Git clone this repo `cd AirQo-api/src/location-registry` . And then run the following docker commands
`docker build -t us.gcr.io/airqo-250220/airqo-location-registry-api .` and `docker run -p 4000:8080 us.gcr.io/airqo-250220/airqo-location-registry-api`
All endpoints as shown [in this documentation](https://docs.google.com/spreadsheets/d/1-NGTDBtTcIi8Fg2UYSWjcK6geKrZdopz5kxRFsCEMPs/edit#gid=1270588885) You just need to add the tenant parameter to each url e.g. http://localhost:4000/api/v1/location_registry/create_id?tenant=kcca

Be sure to get the latest .env file

### DIFF
--- a/src/location-registry/Dockerfile
+++ b/src/location-registry/Dockerfile
@@ -16,6 +16,9 @@ RUN virtualenv /env -p python3.7
 ENV VIRTUAL_ENV /env
 ENV PATH /env/bin:$PATH
 
+#Setting running environment to production
+ENV FLASK_ENV=production
+
 # Copy the application's requirements.txt and run pip to install all
 # dependencies into the virtualenv.
 ADD requirements.txt /app/requirements.txt
--- a/src/location-registry/app.py
+++ b/src/location-registry/app.py
@@ -5,13 +5,15 @@ from dotenv import load_dotenv
 from flask_cors import CORS
 from flask_pymongo import PyMongo
 from controllers.controller import location_blueprint, cache
+from config import app_config
 load_dotenv()
 
 _logger = logging.getLogger(__name__)
 app = Flask(__name__)
+app.config.from_object(app_config["development"])
 CORS(app)
 cache.init_app(app)
-app.config["MONGO_URI"] = os.getenv("MONGO_URI")
+#app.config["MONGO_URI"] = os.getenv("MONGO_URI")
 mongo = PyMongo(app)
 app.register_blueprint(location_blueprint)
 
--- a/src/location-registry/config.py
+++ b/src/location-registry/config.py
@@ -19,7 +19,7 @@ class ProductionConfig(Config):
     load_dotenv(dotenv_path)
     DEBUG = False
     #SERVER_PORT = os.environ.get('PORT', 5000)
-    MONGO_URI = os.getenv("MONGO_URI")
+    MONGO_URI = os.getenv("PROD_MONGO_URI")
 
 
 class DevelopmentConfig(Config):
@@ -28,7 +28,7 @@ class DevelopmentConfig(Config):
 
     DEVELOPMENT = True
     DEBUG = True
-    MONGO_URI = os.getenv("MONGO_URI")
+    MONGO_URI = os.getenv("DEV_MONGO_URI")
 
 
 class TestingConfig(Config):
@@ -36,7 +36,7 @@ class TestingConfig(Config):
     load_dotenv(dotenv_path)
 
     TESTING = True
-    MONGO_URI = os.getenv("MONGO_URI")
+    MONGO_URI = os.getenv("DEV_MONGO_URI")
 
 
 app_config = {"development": DevelopmentConfig,
@@ -44,4 +44,4 @@ app_config = {"development": DevelopmentConfig,
 
 
 if __name__ == '__main__':
-    print('package root', "Locations Registry App")
+    print('package root', "Location Registry App")
--- a/src/location-registry/controllers/controller.py
+++ b/src/location-registry/controllers/controller.py
@@ -33,10 +33,13 @@ def generate_ref():
     Generates a reference id for a new location
     '''
     tenant_id = request.args.get('tenant')
-    if tenant_id:
-        return helper.get_location_ref(tenant_id)
+    if not tenant_id:
+        return {'message': 'Tenant id required', 'success':False}, 400
+    elif tenant_id != 'airqo':
+        return {'message':'Invalid organization', 'success':False}, 400
     else:
-        return {'message': 'Tenant id required'}, 400
+        return helper.get_location_ref(tenant_id) 
+   
 
 
 @location_blueprint.route('/api/v1/location_registry/register', methods=['POST'])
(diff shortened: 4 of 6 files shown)


Baalmart (commented):
Thanks @lilly-muyama 
I am somehow unable to run it using Docker....

- Tried running it using **GET locations** endpoint and was not successful , nothing was happening - was getting service request error as shown in screenshots.
- Also, what is the request body for /create_id [GET]?

image

image

lilly-muyama:
> 
> 
> Thanks @lilly-muyama
> I am somehow unable to run it using Docker....
> 
>     * Tried running it using **GET locations** endpoint and was not successful , nothing was happening - was getting service request error as shown in screenshots.
> 
>     * Also, what is the request body for /create_id [GET]?
> 
> 
> image
> 
> image

There is no request body for the create_id endpoint. I hadn't used docker to test it. Let me check and see.

danielogen (approved):
Working well. Some [near] future recommendations: Adopting the DRY principle and separation of environments [dev, stage and production]

Baalmart (commented):
I am now able to access Google services, thank God! I have also imported the location data for airqo and kcca into their respective databases. 
- Things are working as expected for get location details endpoint, the create_id endpoint
- So we need to do more error handling, the endpoints always return values even without a tenant specified. See screenshot
- Get details of location to be edited always returns the error respomse
- Secondly, getting the same response for different tenants, see screenshots below. LOCATIONS [GET]
- 
image

image

image

lilly-muyama:
> 
> 
> I am now able to access Google services, thank God! I have also imported the location data for airqo and kcca into their respective databases.
> 
>     * Things are working as expected for get location details endpoint, the create_id endpoint
> 
>     * So we need to do more error handling, the endpoints always return values even without a tenant specified. See screenshot
> 
>     * Get details of location to be edited always returns the error respomse
> 
>     * Secondly, getting the same response for different tenants, see screenshots below. LOCATIONS [GET]
> 
> 
> image
> 
> image
> 
> image

All of these have been addressed.

Baalmart (changes requested):
Thanks, the `Get all locations` works as expected, also returns responses fast.

- [ ] the endpoint for create location ID is confusing me, could be missing something - relating it to a GET is confusing. What is the real purpose of this endpoint?

- [ ] Still on the same endpoint [create location id], works well with airqo tenant but brings an error message when a different tenant is used 

- [ ] could we make that error message in JSON (from HTML) in an appropriate format that includes the cause of the error 

- [ ] Error handling, we can also be able to share the exact cause of an error in our response message. Resource: https://flask.palletsprojects.com/en/master/errorhandling/#:~:text=When%20an%20error%20occurs%20in,user%20when%20an%20error%20occurs. 

- [ ] We also need to separate our environments. The reviewers should be able to switch between different environments easily using different run commands perhaps? You could specify the different run commands (configuration objects) in classes and change ENVs at runtime accordingly.
Resources:
1. https://stackoverflow.com/questions/44951244/flask-production-and-development-mode
2. https://kelvinmwinuka.com/organising-flask-configurations/ 

- [ ] For non existent endpoints, could we return informative JSON responses? 

- [ ] Register Location endpoint - error response is not descriptive for non existent tenant - same case for existing tenant name - guess we need to handle cases where the user forgets to provide a request body, give them an appropriate JSON error response perhaps? - this particular endpoint also takes extremely long to finish execution. 

- [ ] Get details of location works as expected. Just that when a user forgets to enter the tenant AND when they enter a non existent tenant ID, they get the same error message reponse, we could try to differentiate the response for those two scenarios using appropriate error status codes and reponses. Same case for non existent location names ---all these three scenarios return the same error response. 

- [ ] For update Location, please change the HTTP verb to PUT instead of POST. Also, could we return the details of the updated location? 

- [ ] Get details of a location to be edited - As for this endpoint, we shall want to consider avoiding this network call - what if all the location details are loaded on user signin using redux and then a location's details are also picked from state. We shall also want to use a more appropriate HTTP verb (PUT) for all scenarios of updating and editing records. Otherwise, the endpoint works as expected. One last thing, what is the difference between this endpoint and "GET details of a location"? Lastly, we need to review the error responses as shared above.

Priscilla-AL (commented):
The view function did not return a valid response. The function either returned None or ended without a return statement.

ja-odur (approved):
Tested all `GET` endpoints and works perfectly

lilly-muyama:
> 
> 
> Thanks, the `Get all locations` works as expected, also returns responses fast.
> 
>     * [ ]  the endpoint for create location ID is confusing me, could be missing something - relating it to a GET is confusing. What is the real purpose of this endpoint?
> 
>     * [ ]  Still on the same endpoint [create location id], works well with airqo tenant but brings an error message when a different tenant is used
> 
>     * [ ]  could we make that error message in JSON (from HTML) in an appropriate format that includes the cause of the error
> 
>     * [ ]  Error handling, we can also be able to share the exact cause of an error in our response message. Resource: https://flask.palletsprojects.com/en/master/errorhandling/#:~:text=When%20an%20error%20occurs%20in,user%20when%20an%20error%20occurs.
> 
>     * [ ]  We also need to separate our environments. The reviewers should be able to switch between different environments easily using different run commands perhaps? You could specify the different run commands (configuration objects) in classes and change ENVs at runtime accordingly.
>       Resources:
> 
> 
>     1. https://stackoverflow.com/questions/44951244/flask-production-and-development-mode
> 
>     2. https://kelvinmwinuka.com/organising-flask-configurations/
> 
> 
>     * [ ]  For non existent endpoints, could we return informative JSON responses?
> 
>     * [ ]  Register Location endpoint - error response is not descriptive for non existent tenant - same case for existing tenant name - guess we need to handle cases where the user forgets to provide a request body, give them an appropriate JSON error response perhaps? - this particular endpoint also takes extremely long to finish execution.
> 
>     * [ ]  Get details of location works as expected. Just that when a user forgets to enter the tenant AND when they enter a non existent tenant ID, they get the same error message reponse, we could try to differentiate the response for those two scenarios using appropriate error status codes and reponses. Same case for non existent location names ---all these three scenarios return the same error response.
> 
>     * [ ]  For update Location, please change the HTTP verb to PUT instead of POST. Also, could we return the details of the updated location?
> 
>     * [ ]  Get details of a location to be edited - As for this endpoint, we shall want to consider avoiding this network call - what if all the location details are loaded on user signin using redux and then a location's details are also picked from state. We shall also want to use a more appropriate HTTP verb (PUT) for all scenarios of updating and editing records. Otherwise, the endpoint works as expected. One last thing, what is the difference between this endpoint and "GET details of a location"? Lastly, we need to review the error responses as shared above.

@Baalmart I have handled most of these where applicable. Please re-review and let me know if you have any additional comments.

Baalmart (approved):
Good

PR #154, opened 2020-09-29
pseudo code for inserting device measurements

Baalmart:
# Topic

Description
Just sharing pseudo code for inserting device measurements
Also highlights the meeting minutes - with some suggestions written as comments

Just check out the changed file and walk through the minutes of meeting with some pseudocode, no need to run anything - just observe.

### DIFF
--- a/src/device-registry/utils/insertion.js
+++ b/src/device-registry/utils/insertion.js
@@ -1,0 +1,233 @@
+/***
+ * the data is in this format
+ */
+const deviceMeasurements = [
+  "created_at",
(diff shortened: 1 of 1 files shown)


Priscilla-AL (commented):
How are the values for frequency and weight being added? They are not part of device measurements

Baalmart:
Hi @Priscilla-AL :
The `frequency` and `weight `are provided in the request body as shown in the attached screenshot:
image


> How are the values for frequency and weight being added? They are not part of device measurements

PR #202, opened 2020-10-29
Bg stream kcca device data

NoahNsimbe:
# Kcca device measurements
**Bug**

**_WHAT DOES THIS PR DO?_**
Fixes a bug that caused duplication of the device measurements

[PLAT-268](https://airqoteam.atlassian.net/browse/PLAT-268)

[bg-stream-kcca-device-data](https://github.com/airqo-platform/AirQo-api/tree/bg-stream-kcca-device-data)

### DIFF
--- a/pipeline/cloud-functions/python/cloud_function_process_kcca_device_measurements.py
+++ b/pipeline/cloud-functions/python/cloud_function_process_kcca_device_measurements.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 import os
 import pandas as pd
 import requests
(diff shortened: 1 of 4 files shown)


pipeline/cloud-functions/python/cloud_function_process_kcca_device_measurements.py
```diff
@@ -36,8 +31,9 @@ def get_kcca_device_data():
     :return: current kcca device measurements
     """
 
-    # get current date : %Y-%m-%d
-    date = convert_date_to_formatted_str(datetime.now(), 'daily')
+    # get current date and time 5 minutes ago : %Y-%m-%dT%H:%M:%SZ
+    # the cron job must be scheduled to run as the time interval stated here
```
Baalmart, line 35: Thanks @NoahNsimbe :
Is this where the magic happens?
Just curious, what informs the 5 minutes interval?
Is this the way we are getting rid of the duplicate values?

NoahNsimbe: @Baalmart : 
I looked at the data from the API and realized there are no duplicate values. The duplications were caused by the way we were fetching the data. Initially, we were querying the API every 5 minutes for data of the whole day and assuming the API will return only new data (data that wasn't returned previously). For every fetch, the API returned already fetched data plus new data which caused the duplication. So fetching data for the last n minutes/hours/days and running the function after every n minutes/hours/days will return new data.

Baalmart: Awesome, thanks.

Issue #358, opened 2021-05-03
user does not exist on password reset with correct token


Baalmart:
**What were you trying to achieve?**
Trying to reset a password using the password reset endpoint

**What are the expected results?**
The password should get updated from email address link accordingly.

**What are the received results?**
```
{
    "success": false,
    "message": "user does not exist"
}
```

**What are the steps to reproduce the issue?**
Just head over to the password reset endpoint and attempt to password reset via email.

**Additional context**
Issues exists in both the backend and the frontend of the system.

Baalmart:
After thorough investigations, the implementation works well in staging environment but does not work well in production environment hence this issue which was tested from production environment.

**What does this mean?** We need to bring the recent changes of auth service staging into production asap. cc: @danielogen

Baalmart:
This was just an issue of **Cast Errors** due to the corrupt data where phone numbers were imported as NaN and yet the system expected a type of NUMBER. This issue prevented this feature from executing perfectly. So the fix has just been more about making manual updates to the DB to replace the NaNs with actual numbers. cc: @danielogen

PR #359, opened 2021-05-03
merge staging into production

Baalmart:
# merge staging into production

**_WHAT DOES THIS PR DO?_**
merge staging into production.


danielogen (approved):
LGTM

Baalmart:
But wait a second. @danielogen , I do not see any file changes for the auth service as to regards to this issue [358](https://github.com/airqo-platform/AirQo-api/issues/358) which is only appearing in the production environment and **not** the staging environment.

So this means that we need to examine this issue more.

danielogen:
> But wait a second. @danielogen , I do not see any file changes for the auth service as to regards to this issue [358](https://github.com/airqo-platform/AirQo-api/issues/358) which is only appearing in the production environment and **not** the staging environment.
> 
> 
> 
> So this means that we need to examine this issue more.

True, I also had the same observation but I thought I might be missing something!

danielogen (approved):
LGTM

PR #384, opened 2021-05-17
Filter active or primary Devices

NoahNsimbe:
# Filter primary and active devices

**_WHAT DOES THIS PR DO?_**
Enhances device registry by adding filtering for devices that are active or primary in a location

[PLAT-550](https://airqoteam.atlassian.net/browse/PLAT-550?atlOrigin=eyJpIjoiOWI3NzY2YWYyOGMwNDA0NTg5YzA4YzBlM2QyZDUxZjMiLCJwIjoiaiJ9)
[PLAT-554](https://airqoteam.atlassian.net/browse/PLAT-554?atlOrigin=eyJpIjoiYzVkZDM3ZjBlMDNiNDc0ZWFhYmFiMDQ0ZmRmMjFjMmEiLCJwIjoiaiJ9)

[filter-by-primary-in-location](https://github.com/airqo-platform/AirQo-api/tree/filter-by-primary-in-location)

checkout this branch
npm install
npm run stage-mac OR npm run stage-pc

http://localhost:3000/api/v1/devices?tenant=airqo&primary=yes&active=yes
http://localhost:3000/api/v1/devices?tenant=airqo&primary=no
http://localhost:3000/api/v1/devices?tenant=airqo&active=yes
http://localhost:3000/api/v1/devices?tenant=airqo&active=no

[Docs](https://docs.airqo.net/airqo-platform-api/device-registry#list-device-s)

None
None

### DIFF
--- a/src/device-registry/utils/generate-filter.js
+++ b/src/device-registry/utils/generate-filter.js
@@ -64,99 +64,73 @@ const generateDeviceFilter = (
   location,
   siteName,
   mapAddress,
-  primary
+  primary,
+  active
 ) => {
-  if (tenant && name && !channel && !location && !siteName && !mapAddress && !primary) {
+
+  if(!tenant){
+    return {}
+  }
+
+  let filter = {};
+
+  if(name){
     let regexExpression = generateRegexExpressionFromStringElement(name);
-    let filter = {
-      name: { $regex: regexExpression, $options: "i" },
-    };
-    return filter;
-  } else if (
-    tenant &&
-    !name &&
-    channel &&
-    !location &&
-    !siteName &&
-    !mapAddress &&
-    !primary
-  ) {
-    return {
-      channelID: channel,
-    };
-  } 
-  
-  else if (
-    tenant &&
-    !name &&
-    !channel &&
-    !location &&
-    !siteName &&
-    !mapAddress &&
-    primary
-  ) { 
+    filter['name'] = { $regex: regexExpression, $options: "i" };
+  }
+
+  if(channel){
+    filter['channelID'] = channel;
+  }
+
+  if(location){
+    filter['locationID'] = location;
+  }
+
+  if(siteName){
+    let regexExpression = generateRegexExpressionFromStringElement(siteName);
+    filter['siteName'] = { $regex: regexExpression, $options: "i" };
+  }
+
+  if(mapAddress){
+    let regexExpression = generateRegexExpressionFromStringElement(mapAddress);
+    filter['locationName'] = { $regex: regexExpression, $options: "i" };
+  }
+
+  if (primary) { 
+
+    const primaryStr = primary + "";
     
-    if(primary == 'yes'){
-      return {
-        isPrimaryInLocation: true,
-      }
+    if(primaryStr.toLowerCase() == 'yes'){
+      filter['isPrimaryInLocation'] = true;
     }
-    else if(primary == 'no'){
-      return {
-        isPrimaryInLocation: false,
-      }
+    else if(primaryStr.toLowerCase() == 'no'){
+      filter['isPrimaryInLocation'] = false;
     }
     else{
-      return {
-  
-      };
+
     }
 
   }
-  
-  else if (
-    tenant &&
-    !name &&
-    !channel &&
-    location &&
-    !siteName &&
-    !mapAddress &&
-    !primary
-  ) {
-    return {
-      locationID: location,
-    };
-  } else if (
-    tenant &&
-    !name &&
-    !channel &&
-    !location &&
-    siteName &&
-    !mapAddress &&
-    !primary
-  ) {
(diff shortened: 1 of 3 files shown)


src/device-registry/utils/get-device-details.js
```diff
@@ -18,6 +18,7 @@ const getDetail = async (
   loc,
```
Baalmart, line 18: We shall need to review this util function to ensure that this change in its parameters does not break anything in any other piece of code which utilises it.

I guess this screams the need to fast track the unit test implementations for this microservice so that any issues are captured automatically and not manually. :)

NoahNsimbe: thanks @Baalmart, enforcing TDD/unit tests for this microservice shall commence next sprint.

src/device-registry/utils/generate-filter.js
```diff
@@ -76,18 +78,50 @@ const generateDeviceFilter = (
...
+    !channel &&
+    !location &&
+    !siteName &&
+    !mapAddress &&
+    primaryInLocation
+  ) { 
+    
+    if(primaryInLocation == 'true'){
+      return {
+        isPrimaryInLocation: true,
```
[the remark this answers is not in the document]
NoahNsimbe: Thanks for the feedback @Baalmart, the query parameter has been changed.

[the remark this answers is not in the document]
NoahNsimbe: @Baalmart, I have added the possible filter combinations

ja-odur (approved):
Tested the endpoints, works as expected

sserurich (approved):
endpoints work as expected

src/device-registry/utils/generate-filter.js
```diff
@@ -62,65 +63,74 @@ const generateDeviceFilter = (
...
-  } else if (tenant && !name && !channel && !location) {
-    return {};
+    filter['locationName'] = { $regex: regexExpression, $options: "i" };
+  }
+
+  if (primary) { 
+
+    const primaryStr = primary + "";
+    
+    if(primaryStr.toLowerCase() == 'yes'){
```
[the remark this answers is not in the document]
NoahNsimbe: @sserurich, the change was made following @Baalmart comments https://github.com/airqo-platform/AirQo-api/pull/384#discussion_r633295302

........ Using Boolean values could potentially bring about some runtime and/or logical errors during operation. " ....... 
"Because of the way we use "truthiness" when checking for presence of a value in JS, I suggest that we explore using values/strings like "yes" and "no" when checking the values for the query parameter. "

src/device-registry/utils/generate-filter.js
```diff
@@ -62,65 +63,74 @@ const generateDeviceFilter = (
...
   location,
   siteName,
-  mapAddress
+  mapAddress,
+  primary,
+  active
 ) => {
-  if (tenant && name && !channel && !location && !siteName && !mapAddress) {
+
+  if(!tenant){
```
Baalmart, line 71: Thanks @NoahNsimbe 
We shall need to review this logic here.
In case it is only the tenant provided, then we need to generate an empty filter.
Hence something along the lines of:
```
if (tenant){
return {}
}
```
Or better still, employ multiple else if statements and end with the default one which returns an empty object. More like:
```
if (name){
filter['name'] = { $regex: regexExpression, $options: "i" };
}
else if (channel){
filter['channelID'] = channel;
}
else if (location){
filter['locationID'] = location;
}
.....etc
else {
return {}
}
```

Actually, tenant is not required in this use case for generating filters.

We also need to review all possible filter combinations in the process. Example: what if one provides both the location and  the channel in the query parameters? What kind of filter shall we have then? I assume the filter there would be: 
```
{
channelID: channel,
locationID: location
}
```
This will then lead us to review the way the assignments are being done throughout this filtration process. 
In summary, we need to review all possible query combinations and test them out accordingly.

NoahNsimbe: @Baalmart, I have used `if` statements, meaning that if somebody provides for than one query parameter, they are all captured, for example checkout the following in the code. With approach, I believe a query filter is captured as long as it has been provided because execution continues till the return statement. The logic is that if a filter is provided, it is added to the `filter` variable and execution continues to check for other filters, if the filter is not provided, still execution continues and checks for other filters
```
 if(channel)
{    filter['channelID'] = channel;  }  
if(location)
{    filter['locationID'] = location;  }
```

Even if they provide only the tenant, `{}` is returned. Checkout `let filter = {};` in the code that is returned at the end.

Baalmart: Makes sense @NoahNsimbe , thanks for the clarification.

Baalmart (approved):
All LGTM now, thanks @NoahNsimbe

PR #441, opened 2021-06-22
Streams multi tenancy

NoahNsimbe:
# Streams Multi tenancy

**_WHAT DOES THIS PR DO?_**
Enhances the streaming application to include multi-tenancy

**_WHAT JIRA STORY/TASK IS RELATED TO THIS PR?_**
[PLAT-632](https://airqoteam.atlassian.net/browse/PLAT-632?atlOrigin=eyJpIjoiZjI0ZGM4MzM0NzEyNDhmODlmZDU3MmVkYmRmNTQyNjMiLCJwIjoiaiJ9)
[PLAT-631](https://airqoteam.atlassian.net/browse/PLAT-631?atlOrigin=eyJpIjoiZDI1MzcwMmRhMzA5NDEyZDg0MWMwMzU2OGNmZWRhYTciLCJwIjoiaiJ9)

**_WHAT IS THE LINK TO THE PR BRANCH_**
[streams-multi-tenancy](https://github.com/airqo-platform/AirQo-api/tree/streams-multi-tenancy)

**_HOW DO I TEST OUT THIS PR?_**
- You must have [Kafka](https://kafka.apache.org/quickstart) and a [schema registry](https://github.com/airqo-platform/AirQo-api/pull/411) setup

#### Create topics to store the measurements
```
bin/kafka-topics.sh --create --topic transformed-device-measurements-topic --bootstrap-server localhost:9092
bin/kafka-topics.sh --create --topic kcca-raw-device-measurements-topic --bootstrap-server localhost:9092
bin/kafka-topics.sh --create --topic airqo-raw-device-measurements-topic --bootstrap-server localhost:9092
```

#### Run the application and view its logs
```
cd AirQo-api/src/data-mgt/java/streams/device-measurements
```
Update the **bootstrap.servers** value in the application.properties file under `AirQo-api/src/data-mgt/java/streams/device-measurements/src/main/resources`  to **bootstrap.servers=localhost:9092**
```
docker build -t kafka-kcca-measurements-stream .
docker run --network="host" --rm kafka-kcca-measurements-stream
```

#### Using a different terminal, view the device measurements on the  transformed-device-measurements-topic topic. You should see bytes
```
bin/kafka-console-consumer.sh --topic transformed-device-measurements-topic --from-beginning --bootstrap-server localhost:9092
```

**_WHICH ENDPOINTS SHOULD BE READY FOR TESTING?:_** 
None

[...]

PR #440, opened 2021-06-21
Kcca frequency measurements

NoahNsimbe:
# Kcca meaurements connector

**_WHAT DOES THIS PR DO?_**
Creates a kafka connector to  fetch and store Kcca frequency device measurements on a topic

**_WHAT JIRA STORY/TASK IS RELATED TO THIS PR?_**
[Stream all Kcca frequency measurements (hour and day)](https://airqoteam.atlassian.net/browse/PLAT-632?atlOrigin=eyJpIjoiMTgxNTc1Mjg0NTQ3NDIyN2FkYTQ0NWIxMTdhMjI4YTUiLCJwIjoiaiJ9)

**_WHAT IS THE LINK TO THE PR BRANCH_**
[ft-kcca-frequency-measurements](https://github.com/airqo-platform/AirQo-api/tree/ft-kcca-frequency-measurements)

**_HOW DO I TEST OUT THIS PR?_**
- You must have access to a kubernetes cluster with kafka installed using strimzi
- https://strimzi.io/quickstarts/

 **_Wait for the cluster to be created._**

 **_Wait for the cluster to be created._**

**_WHICH ENDPOINTS SHOULD BE READY FOR TESTING?:_** 
None

**_IS THERE ANY JENKINS CONSOLE LINK FOR CODE COVERAGE AND BUILD INFO?_**
None

**_ARE THERE ANY RELATED PRs?_**
None

PR #412, opened 2021-06-14
Airqo meaurements connector

NoahNsimbe:
# Airqo meaurements connector

**_WHAT DOES THIS PR DO?_**
Creates a kafka connector to  fetch and store AirQo device measurements on a topic

**_WHAT JIRA STORY/TASK IS RELATED TO THIS PR?_**
None

**_WHAT IS THE LINK TO THE PR BRANCH_**
[airqo-meaurements-connector](https://github.com/airqo-platform/AirQo-api/tree/airqo-meaurements-connector)

**_HOW DO I TEST OUT THIS PR?_**
- You must have access to a kubernetes cluster with kafka installed using strimzi
- https://strimzi.io/quickstarts/

 **_Wait for the cluster to be created._**

 **_Wait for the cluster to be created._**

**_WHICH ENDPOINTS SHOULD BE READY FOR TESTING?:_** 
None

**_IS THERE ANY JENKINS CONSOLE LINK FOR CODE COVERAGE AND BUILD INFO?_**
None

**_ARE THERE ANY RELATED PRs?_**
None

PR #411, opened 2021-06-14
Schema registry

NoahNsimbe:
# Schema registry

**_WHAT DOES THIS PR DO?_**
Introduces Schema Registry as a storage location for the format of messages exchanged with the message broker(Kafka)

**_WHAT JIRA STORY/TASK IS RELATED TO THIS PR?_**
[Setup a schema registry for Kafka](https://airqoteam.atlassian.net/browse/PLAT-631?atlOrigin=eyJpIjoiMDgxOGFiMDI3ODc0NDc2ODg4OWIxNjZhNmVlYTFlODYiLCJwIjoiaiJ9)

**_WHAT IS THE LINK TO THE PR BRANCH_**
[ft-schema-registry](https://github.com/airqo-platform/AirQo-api/tree/ft-schema-registry)

**_HOW DO I TEST OUT THIS PR?_**
Follow instructions in the `Readme.md` file under `AirQo-api/kafka/schemas`. The easiest way is to test using Docker. Ports `9092` and `8081` must be available.

**_WHICH ENDPOINTS SHOULD BE READY FOR TESTING?:_** 
None

**_IS THERE ANY JENKINS CONSOLE LINK FOR CODE COVERAGE AND BUILD INFO?_**
None

**_ARE THERE ANY RELATED PRs?_**
#441

PR #443, opened 2021-06-22
Schema registry <> Device registry

NoahNsimbe:
# Using Schema Registry for message deserialization.

**_WHAT DOES THIS PR DO?_**
Updates device registry to use schema registry for message deserialization.

[PLAT-631](https://airqoteam.atlassian.net/browse/PLAT-631?atlOrigin=eyJpIjoiNDA2ZDBlM2RlMzY5NGQ5ODk5M2UxYTVjNGQ4ZjM0MzMiLCJwIjoiaiJ9)

[schema-registry-device-registry](https://github.com/airqo-platform/AirQo-api/tree/schema-registry-device-registry)

- You must have [Kafka](https://kafka.apache.org/quickstart),  [schema registry](https://github.com/airqo-platform/AirQo-api/pull/411) and a [streaming application](https://github.com/airqo-platform/AirQo-api/pull/441) setup.
- The latest [.env file](https://docs.google.com/document/d/1R0ux6i4WaK-NssVQE_9IcTxfUs3ZYIUhSP1Aa7s_hqg/edit?usp=sharing)

```
npm install
```
`npm run dev-mac` and `npm run dev-pc` for Windows

[Get Events](https://docs.airqo.net/airqo-platform-api/device-registry#get-events)

None

#411 
#441 
#412 
#440


src/device-registry/controllers/kafka/raw-device-measurements-consumer.js
```diff
@@ -1,52 +1,32 @@
-const kafka = require('kafka-node');
-const Consumer = kafka.Consumer;
 const { logObject, logElement } = require("../../utils/log");
```
Baalmart, line 1: Could we try to avoid having files in many subfolders?
Having a near flat code organisational structure is good for many things, including the ease of maintenance + unit testing components. @NoahNsimbe

src/device-registry/config/constants.js
```diff
@@ -7,6 +7,8 @@ const devConfig = {
   REDIS_PORT: process.env.REDIS_PORT,
   KAFKA_BOOTSTRAP_SERVERS: process.env.KAFKA_BOOTSTRAP_SERVERS_DEV,
   KAFKA_TOPICS: process.env.KAFKA_TOPICS_DEV,
```
Baalmart, line 9: looks neat.

src/device-registry/utils/insert-device-measurements.js
```diff
@@ -1,54 +1,50 @@
...
+                logObject("Kafka Measurement", measurement);
 
-                const device = data.device;
-                const tenant = data.tenant;
-                const measurements = [data];
+                const device = measurement.device;
+                const tenant = measurement.tenant;
+                const measurements = [measurement];
 
                 if (!tenant || !device) {
```
Baalmart, line 22: Could you consider using && here?
Why? With the | | operator, it will short circuit the comparison and accept cases where !tenant is true even the !device is false at the same time. Please review the use case and advise the intention.

Baalmart (changes requested):
All looks great, just a few review comments for your review @NoahNsimbe

src/device-registry/controllers/kafka/raw-device-measurements-consumer.js
NoahNsimbe: thanks @Baalmart I am going to make the change....The motivation was to create a clear boundary between REST API logic and Message Broker logic because both provide exist and entry points for the microservice in totally different ways. 
 **1.** This could make it easy to identify and test message broker logic as well as the rest api logic. 
**2.** Making changes to the message broker shouldn't involve making changes to the rest api which makes maintenance, code update and review easy since a developer knows the folder to mess around with. In the future, if the microservice happens to have an embedded web interface for admin purposes then creating folders could be of great importance when the question "what do you want to change/update ??...is it the admin web interface, api endpoints or message broker connections??" is asked

src/device-registry/utils/insert-device-measurements.js
NoahNsimbe: thanks @Baalmart the right operator is **&&**, I didn't think about that earlier

NoahNsimbe: @Baalmart the logic is valid basing on the intention...if device or tenant is missing, reject the measurements

src/device-registry/controllers/kafka/raw-device-measurements-consumer.js
Baalmart: Thanks, all makes sense @NoahNsimbe ....was just worried about deeply nested files/directories. They usually becomes a maintenance nightmare since one has to navigate many subdirectories to reach their destination. You can read more about how to avoid deeply nested directory structures especially in our java code: https://stackoverflow.com/questions/26131352/avoid-deep-directory-structure-in-a-java-project

src/device-registry/utils/insert-device-measurements.js
Baalmart: Cool, thanks.

Baalmart: Hi @NoahNsimbe , the tenant is always required.....in ALL situations. There is not scenario (usecase) that is possible without a tenant. Please crosscheck further.

Baalmart (changes requested):
Most of it LGTM, thanks @NoahNsimbe.  Just some pending issue.......
Please review my latest review comment and make appropriate changes. Thanks.

src/device-registry/utils/insert-device-measurements.js
NoahNsimbe: @Baalmart what is your suggestion?

Baalmart: Hi @NoahNsimbe , there was an earlier suggestion (&&) in this thread. But I guess in this scenario, the logic is fine because the response body is not specific. Thanks! :)

Baalmart (approved):
LGTM now....thanks @NoahNsimbe , good stuff!

danielogen (approved):
LGTM

PR #458, opened 2021-07-06
Kafka device registry

NoahNsimbe:
# <Feature/Chore/Hotfix/Topic/Bug Title>

**_WHAT DOES THIS PR DO?_**
Creates a separate deployment for kafka on device registry

None

[kafka-device-registry](https://github.com/airqo-platform/AirQo-api/tree/kafka-device-registry)

```
npm install
npm run kafka-stage-mac
```
None

None

None

### DIFF
--- a/src/device-registry/bin/kafka-connect.js
+++ b/src/device-registry/bin/kafka-connect.js
@@ -1,8 +1,8 @@
 const dotenv = require("dotenv");
 dotenv.config();
 
-const { kafkaMeasurementsConsumer } = require("../controllers/raw-device-measurements-consumer");
+const { rawEventsConsumer } = require("../controllers/events-consumer");
 const { mongodb } = require("../config/database");
 
 mongodb;
-kafkaMeasurementsConsumer;
+rawEventsConsumer;
--- a/src/device-registry/bin/kafka-connect.js
+++ b/src/device-registry/bin/kafka-connect.js
@@ -1,0 +1,8 @@
+const dotenv = require("dotenv");
+dotenv.config();
+
+const { kafkaMeasurementsConsumer } = require("../controllers/raw-device-measurements-consumer");
+const { mongodb } = require("../config/database");
+
+mongodb;
+kafkaMeasurementsConsumer;
--- a/src/device-registry/bin/www
+++ b/src/device-registry/bin/www
@@ -10,7 +10,6 @@ const middlewareConfig = require("../config/app.middleware");
 var debug = require("debug")("device-registry:server");
 var http = require("http");
 var isEmpty = require("is-empty");
-const { kafkaMeasurementsConsumer } = require("../controllers/raw-device-measurements-consumer");
 
 /****
  * pass the app instance to the middleware configuration
@@ -105,5 +104,4 @@ Running on ${ENV}`);
     var addr = server.address();
     var bind = typeof addr === "string" ? "pipe " + addr : "port " + addr.port;
     debug("Listening on " + bind);
-    // kafkaMeasurementsConsumer;
 }
--- a/src/device-registry/controllers/events-consumer.js
+++ b/src/device-registry/controllers/events-consumer.js
@@ -1,7 +1,7 @@
 const { logObject, logElement } = require("../utils/log");
 const { Kafka } = require('kafkajs')
 const { SchemaRegistry } = require('@kafkajs/confluent-schema-registry')
-const insertMeasurtements = require("../utils/insert-device-measurements");
+const insertMeasurtements = require("../utils/bulk-insert-measurements");
 const constants = require("../config/constants");
 const SCHEMA_REGISTRY = constants.SCHEMA_REGISTRY;
 const BOOTSTRAP_SERVERS = constants.KAFKA_BOOTSTRAP_SERVERS;
@@ -13,7 +13,7 @@ const kafka = new Kafka({ clientId: KAFKA_CLIENT_ID, brokers: [BOOTSTRAP_SERVERS
 const registry = new SchemaRegistry({ host: SCHEMA_REGISTRY })
 const consumer = kafka.consumer({ groupId: KAFKA_CLIENT_GROUP })
 
-const rawMeasurementsConsumer = async () => {
+const rawEventsConsumer = async () => {
   await consumer.connect();
 
   const topics = RAW_MEASUREMENTS_TOPICS.split(",");
@@ -36,6 +36,6 @@ const rawMeasurementsConsumer = async () => {
   });
 }
 
-rawMeasurementsConsumer().catch(console.error)
+rawEventsConsumer().catch(console.error)
 
-module.exports = rawMeasurementsConsumer;
+module.exports = rawEventsConsumer;
--- a/src/device-registry/package.json
+++ b/src/device-registry/package.json
@@ -12,7 +12,10 @@
         "prod-pc": "SET NODE_ENV=production&&nodemon ./bin/www",
         "stage-pc": "SET NODE_ENV=staging&&nodemon ./bin/www",
         "prettier": "prettier --single-quote --print-width 80 --trailing-comma all --write 'src/**/*.js'",
-        "test": "mocha ./**/test/ut_*.js --exit"
+        "test": "mocha ./**/test/ut_*.js --exit",
+        "kafka-dev-mac": "NODE_ENV=development nodemon ./bin/kafka-connect.js",
+        "kafka-stage-mac": "NODE_ENV=staging nodemon ./bin/kafka-connect.js",
+        "kafka-prod-mac": "NODE_ENV=production nodemon ./bin/kafka-connect.js"
     },
     "dependencies": {
         "@google-cloud/bigquery": "^5.5.0",
--- a/src/device-registry/utils/bulk-insert-measurements.js
+++ b/src/device-registry/utils/bulk-insert-measurements.js
@@ -5,8 +5,6 @@ const { logObject } = require("./log");
 
 const insertDeviceMeasurements = {
 
-    // Contains modules that provide various ways to insert device measurements.
-
     addValuesArray: async (measurements_array) => {
 
         try {
@@ -73,75 +71,6 @@ const insertDeviceMeasurements = {
 
     },
 
-    addDeviceValues: async (device, tenant, measurements) => {
-
-        try {
-
-            if (tenant && device && measurements) {
-
-                const deviceDetails = await getDetail(tenant, device);
-                const deviceExists = isEmpty(deviceDetails);
-    
-                if (!deviceExists){
-                    return {
-                        success: false,
-                        message: `Device (${device}) for tenant (${tenant}) does not exist on the network`,
-                        errors: [],
-                        valuesRejected: data,
-                        valuesAdded: [],
-                    }
-                }
-
-                const transformedMeasurements = await transformMeasurements(
-                    device,
-                    measurements
-                  );
-
-                const response = await insertMeasurements(
-                    tenant,
-                    transformedMeasurements
-                );
-
-                if(response.success === false){
-                    return {
-                        success: false,
-                        message: "finished the operation with some errors",
-                        errors: response.errors,
-                        valuesRejected: response.valuesRejected,
-                        valuesAdded: response.valuesAdded,
-                      };
-                }
-
-                return response;
-
-
-            } else {
-
-                return {
-                    success: false,
-                    message: "misssing parameters",
-                    errors: [],
-                    valuesRejected: data,
-                    valuesAdded: [],
-
-                };
-            }
-
-
-        } 
-        catch (e) {
-
-            return {
-                success: false,
-                message: "Error Occurred...",
-                errors: error.message,
-                valuesRejected: [],
-                valuesAdded: [],
-            };
-
-        }
-
-    },
 };
 
 module.exports =  insertDeviceMeasurements;
--- a/src/device-registry/Dockerfile.kafka.stage
+++ b/src/device-registry/Dockerfile.kafka.stage
@@ -1,0 +1,11 @@
+FROM node:12
+
+WORKDIR /usr/src/app
+
+COPY package*.json ./
+
+RUN npm install
+
+COPY . .
+
+CMD [ "npm", "run", "kafka-stage-mac" ]


src/device-registry/controllers/events-consumer.js
```diff
@@ -0,0 +1,41 @@
+const { logObject, logElement } = require("../utils/log");
```
Baalmart, line 1: this looks good, like that we have reduced the layers of folders involved in reaching this controller.

src/device-registry/package.json
```diff
@@ -12,7 +12,10 @@
         "prod-pc": "SET NODE_ENV=production&&nodemon ./bin/www",
         "stage-pc": "SET NODE_ENV=staging&&nodemon ./bin/www",
         "prettier": "prettier --single-quote --print-width 80 --trailing-comma all --write 'src/**/*.js'",
-        "test": "mocha ./**/test/ut_*.js --exit"
+        "test": "mocha ./**/test/ut_*.js --exit",
+        "kafka-dev-mac": "NODE_ENV=development nodemon ./bin/kafka-connect.js",
```
Baalmart, line 16: Yeah, this looks great. I gather that in the near future, we shall migrate this to a new microservice or just embed this inside the data-mgt/node microservice.

src/device-registry/bin/www
```diff
@@ -10,7 +10,6 @@ const middlewareConfig = require("../config/app.middleware");
 var debug = require("debug")("device-registry:server");
 var http = require("http");
 var isEmpty = require("is-empty");
-const { kafkaMeasurementsConsumer } = require("../controllers/kafka/raw-device-measurements-consumer");
```
Baalmart, line 13: Awesome change!

Baalmart (approved):
This LGTM @NoahNsimbe , I like the fact that this will be deployed separately (separate start scripts, etc)----with its own Docker file. :) Great approach!

Please review some of my comments about suggestions/thoughts for future considerations when it comes to file/code organisations in our mono-repo.
Thanks!

Baalmart:
@NoahNsimbe 
Could we also use the same PR to include the corresponding new k8s file? under the k8s directory.....? cc:  @danielogen

Baalmart (changes requested):
@NoahNsimbe
A quick one, could we have these k8s deployment files under the k8s directory.....just to maintain the original flow of things? @danielogen 

Also, I suggest that we introduce the new Kafka changes using this PR. I am extremely hesitant about reversing the the most recent revert.

But let us engage further on the side, just leaving comments here for easy reference....

src/device-registry/controllers/events-consumer.js
NoahNsimbe: 👊

NoahNsimbe:
> @NoahNsimbe
> A quick one, could we have these k8s deployment files under the k8s directory.....just to maintain the original flow of things? @danielogen
> 
> Also, I suggest that we introduce the new Kafka changes using this PR. I am extremely hesitant about reversing the the most recent revert.
> 
> But let us engage further on the side, just leaving comments here for easy reference....

@Baalmart the corresponding device registry k8s deployment file shall be updated. As for the new Kafka changes, including them here will make reviewing very difficult as it will be a mixture of kafka k8s files, java and node. I have a separate PR #462 that handles everything to do with kafka

src/device-registry/stage-docker-build.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+echo -e "\n ==== 🛠  🔧 🔩 Building API docker image 🔩  🔧  🛠  ==== \n"
```
NoahNsimbe, line 3: cc @danielogen

src/device-registry/stage-docker-push.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+echo -e "\n ==== 🚚  🚁 Pushing API docker image 🛫   ==== \n"
+echo -e "\nHope you have built airqo-stage-device-registry-kafka-connection:latest\n"
```
NoahNsimbe, line 4: cc @danielogen

src/device-registry/package.json
NoahNsimbe: Yeah, but it might cause code duplication. Here kafka is acting as an alternative to the insert measurements endpoint on device registry. The only difference is that instead of exposing an endpoint to accept measurements, it gets the measurements from a kafka topic, all the underlying validation and logic is the same. This means device registry has two entry points, kafka and the insert measurements endpoint. Attending to migrate kafka to the data-mgt folder means you have to copy the corresponding logic for inserting measurements and update it whenever you update device registry

Baalmart: Cool, sounds reasonable...thanks. @NoahNsimbe

src/device-registry/utils/bulk-insert-measurements.js
```diff
@@ -5,9 +5,7 @@ const { logObject } = require("../utils/log");
 
 const insertDeviceMeasurements = {
 
-    // Contains modules that provide various ways to insert device measurements.
-    
-    addValuesArray: async (array_data) => {
+    addValuesArray: async (measurements_array) => {
```
Baalmart, line 8: Not sure if it is connected to this but recently realised that the endpoint for inserting many measurements can actually insert values for different devices at the same time....the logic in the code accepts that....the device query parameter provided on the endpoint is secondary in priority so if any measurement in the array has a device attribute, that takes precedence over the query parameter.... :)

Baalmart (commented):
Thanks, this is awesome!
I have just left some more commments @NoahNsimbe

Baalmart (commented):
@NoahNsimbe    A quick one...there are a couple of missing modules despite the run of npm install....so they are bringing about runtime errors.
I notice that your package.json was not changed in this PR, could you reinstall the packages in the project/microservice? Please see attached screenshots:
Screenshot 2021-07-13 at 17 19 36
Screenshot 2021-07-13 at 17 20 23

src/device-registry/utils/bulk-insert-measurements.js
NoahNsimbe: Yes it is connected, But it also depends on the flow of code execution while inserting measurements. Revisit the following files....You might need to make updates where necessary
https://github.com/airqo-platform/AirQo-api/blob/staging/src/device-registry/controllers/create-event.js
https://github.com/airqo-platform/AirQo-api/blob/staging/src/device-registry/utils/transform-measurements.js
https://github.com/airqo-platform/AirQo-api/blob/staging/src/device-registry/utils/insert-measurements.js

NoahNsimbe:
> @NoahNsimbe A quick one...there are a couple of missing modules despite the run of npm install....so they are bringing about runtime errors.
> I notice that your package.json was not changed in this PR, could you reinstall the packages in the project/microservice? Please see attached screenshots:
> Screenshot 2021-07-13 at 17 19 36
> Screenshot 2021-07-13 at 17 20 23

Thanks, I have updated the respective files

Baalmart:
> > @NoahNsimbe
> > A quick one, could we have these k8s deployment files under the k8s directory.....just to maintain the original flow of things? @danielogen
> > Also, I suggest that we introduce the new Kafka changes using this PR. I am extremely hesitant about reversing the the most recent revert.
> > But let us engage further on the side, just leaving comments here for easy reference....
> 
> @Baalmart the corresponding device registry k8s deployment file shall be updated. As for the new Kafka changes, including them here will make reviewing very difficult as it will be a mixture of kafka k8s files, java and nodejs. I have a separate PR #462 that handles everything to do with kafka

Okay, sounds reasonable @NoahNsimbe , thanks.

src/device-registry/utils/bulk-insert-measurements.js
Baalmart: Cool, thanks for the clarification...

src/device-registry/README.md
```diff
@@ -38,10 +39,21 @@ Build the image within the microservice's directory
 docker build -t {IMAGE_NAME} .
 ```
 Then run the container based on the newly created image
-
 ```bash
 docker run -d -n {CONTAINER_NAME} -p host-port:container-port {IMAGE_NAME}
 ```
+To build all images
+```bash
```
Baalmart, line 46: This looks interesting, thanks! :) 
What is the motivation for these bash scripts?
Under what scenario would one want to build all images at once?

src/device-registry/bin/kafka-connect.js
```diff
@@ -0,0 +1,8 @@
+const dotenv = require("dotenv");
+dotenv.config();
```
Baalmart, line 2: In case this is supposed to be a bash script, please consider removing the file extension and adding a[ shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) at the top of the file accordingly

src/device-registry/config/constants.js
```diff
@@ -7,6 +7,10 @@ const devConfig = {
   REDIS_PORT: process.env.REDIS_PORT,
   KAFKA_BOOTSTRAP_SERVERS: process.env.KAFKA_BOOTSTRAP_SERVERS_DEV,
   KAFKA_TOPICS: process.env.KAFKA_TOPICS_DEV,
+  SCHEMA_REGISTRY: process.env.SCHEMA_REGISTRY_DEV,
```
Baalmart, line 10: Awesome stuff! Love that we are following the convention of uppercase while separating the ENVs accordingly....
For items which have the same values in all ENVs, please consider using the defaultConfig (ALONE) below...thanks

src/device-registry/stage-docker-build.sh
Baalmart: Thanks, still trying to appreciate the benefit of having the build these images at the same time? 

- Isn't this introducing some tight coupling of the Kafka and device registry image creation?
- Cant some one create a new image on Kafka without changing the rest of the device registry logic

src/device-registry/utils/bulk-insert-measurements.js
```diff
@@ -0,0 +1,155 @@
+// const { transformMeasurements, bulkTransformMeasurements } = require("../utils/transform-measurements");
+// const { bulkInsert } = require("../utils/insert-measurements");
+// const { doesDeviceExist, doDevicesExist } = require("../utils/does-component-exist");
```
Baalmart, line 3: Could we do away with commented out code from the repository?
By the time, we comment out code, it is likely that we no longer need it so it just pollutes our codebase.
In fact, we need to employ some static analysis tools asap to capture these types of things as we code on our machines.
Statics analysis tools we could use take advantage of include:

- https://codeclimate.com/
- https://www.sonarqube.org/
- https://eslint.org/docs/user-guide/configuring/

src/device-registry/utils/does-component-exist.js
```diff
@@ -48,6 +48,47 @@ const doesDeviceExist = async (deviceName, tenant) => {
   }
 };
 
+const filterExistingDevices = (measurements) => {
+  let promises = measurements.filter(async (measurement) => {
+    try {
+      logText(".......................................");
```
Baalmart, line 54: In the near future, we might actually not need these utils, all we have to do is call the list devices method of the Device entity and proceed accordingly based on the result/response we get. 
More like: 
```

 const responseFromListDevice = await getModelByTenant(
        tenant,
        "device",
        DeviceSchema
      ).list({ name:deviceName})

if (responseFromListDevice.success===false){
// respond to the bad response/error accordingly
}
else {
// go ahead and perform winning actions, all is good...
}
```

src/device-registry/utils/insert-measurements.js
```diff
@@ -112,4 +112,72 @@ const insert = async (tenant, transformedMeasurements) => {
   }
 };
 
-module.exports = insert;
+const bulkInsert = async (transformedMeasurements) => {
```
Baalmart, line 115: What is the difference of this method from the existing one of insert?
Aren't we just duplicating code?
@NoahNsimbe

Baalmart (commented):
Awesome work @NoahNsimbe , things are taking great shape!
Just left some few comments/thoughts for your review.

src/device-registry/bin/kafka-connect.js
NoahNsimbe: @Baalmart this has been fixed

src/device-registry/README.md
NoahNsimbe: When peforming deployment operations

src/device-registry/config/constants.js
NoahNsimbe: @Baalmart, all necessary changes have been made

src/device-registry/utils/insert-measurements.js
NoahNsimbe: The difference is in the method arguments. I have modified the method to call the existing one. @Baalmart

src/device-registry/stage-docker-build.sh
NoahNsimbe: @Baalmart the kafka connection is part of the device registry microservice just like the way the API endpoints are part of the microservice. Changing the underlying logic of the microservice requires building both images.

src/device-registry/stage-docker-build.sh
```diff
@@ -0,0 +1,17 @@
...
+echo "💪 Build complete API docker image"
+echo "💫 Image : airqo-stage-device-registry-api:latest"
+echo "🛫 Push the image : docker push airqo-stage-device-registry-api:latest"
+
+echo -e "\n ==== 🛠  🔧 🔩 Building Kafka connections docker image 🔩  🔧  🛠  ==== \n"
+
+# docker build -t airqo-stage-device-registry-kafka-connection:latest -f Dockerfile.kafka.stage .
+
+echo "💪 Build complete Kafka connections docker image"
+echo "💫 Image : airqo-stage-device-registry-kafka-connection:latest"
```
danielogen, line 16: The image name specified the k8s yaml file is `airqo-stage-device-registry-kafka:latest` cc: @NoahNsimbe

src/device-registry/stage-docker-push.sh
```diff
@@ -0,0 +1,17 @@
...
+echo "💪 Push complete"
+echo "💫 Image : airqo-stage-device-registry-api:latest"
+
+echo -e "\n ==== 🚚  🚁 Pushing Kafka connections docker image 🛫   ==== \n"
+echo -e "\nHope you have built airqo-stage-device-registry-kafka-connection:latest\n"
+
+# docker push airqo-stage-device-registry-kafka-connection:latest
+
+echo "💪 Push complete"
+echo -e "💫 Image : airqo-stage-device-registry-kafka-connection:latest \n"
```
danielogen, line 17: The image name specified the k8s yaml file is `airqo-stage-device-registry-kafka:latest`. @NoahNsimbe

src/device-registry/stage-docker-build.sh
NoahNsimbe: thanks, I have made the updates

src/device-registry/stage-docker-push.sh
NoahNsimbe: 👍, I have made the changes

PR #459, opened 2021-07-06
Revert "Revert "Schema registry <> Device registry""

NoahNsimbe:
Reverts airqo-platform/AirQo-api#455


kafka/clusters/prod-kafka-cluster.yaml
```diff
@@ -1,12 +1,18 @@
-apiVersion: kafka.strimzi.io/v1beta1
```
Baalmart, line 1: @NoahNsimbe 
A quick one, could we have these k8s deployment files under the k8s directory.....just to maintain the original flow of things? @danielogen

Baalmart (changes requested):
just realised something that might require some changes.....awaiting thoughts/concerns...please see the recent comment about the location of the deployment files. @danielogen

kafka/clusters/prod-kafka-cluster.yaml
Baalmart: I guess this comment was intended to the corresponding PR which is a work in progress.......let me migrate it there shortly......I am of the view that we do not revert but rather concentrate on using the other corresponding PR to bring in new changes accordingly. Thanks.

NoahNsimbe: thanks @Baalmart , the requested changes have been worked on in #462

NoahNsimbe:
This pull request has been substituted with #462 and #458

PR #604, opened 2021-09-23
adding projections for different frequencies

Baalmart:
# adding projections for different frequencies

**_WHAT DOES THIS PR DO?_**
adding projections for different frequencies

https://github.com/airqo-platform/AirQo-api/tree/hotfix-create-events

### DIFF
--- a/src/device-registry/models/Event.js
+++ b/src/device-registry/models/Event.js
@@ -335,7 +335,7 @@ eventSchema.statics = {
     }
   },
   list({ skipInt = 0, limitInt = 100, filter = {} } = {}) {
-    const { metadata } = filter;
+    const { metadata, frequency } = filter;
     let search = filter;
     let groupId = "$device";
     let localField = "device";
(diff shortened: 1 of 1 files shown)


NoahNsimbe:
Since we are averaging at API level, how are we dealing with situations where one sensor is malfunctioned or has a null value ??? @Baalmart 
#510

NoahNsimbe (commented):
If somebody needs the values for both sensors, how do they get them?

NoahNsimbe (changes requested):
The fields `average_pm2_5` and `average_pm10`, which are needed in the production version of the app aren't returned when I querying hourly measurements.  I suggest we consider this use case.
https://github.com/airqo-platform/AirQo-frontend/blob/staging/mobile-app/lib/models/measurement.dart

Baalmart:
> The fields `average_pm2_5` and `average_pm10`, which are needed in the production version of the app aren't returned when I querying hourly measurements. I suggest we consider this use case. https://github.com/airqo-platform/AirQo-frontend/blob/staging/mobile-app/lib/models/measurement.dart

Yes, we can get the other fields by using the external query parameter called "external". It can either be "yes" or "no". The default is "yes". So if external is "yes", the response does not include many of the "internally used" fields. @NoahNsimbe

Baalmart:
> If somebody needs the values for both sensors, how do they get them?

Yes, we can get the other fields by using the external query parameter called "external". It can either be "yes" or "no". The default is "yes". So if external is "yes", the response does not include many of the "internally used" fields. @NoahNsimbe

Baalmart:
> Since we are averaging at API level, how are we dealing with situations where one sensor is malfunctioned or has a null value ??? @Baalmart #510

No, we are not doing any manipulation at the API level. All that we do at the API level is change the `presentation` of the data that is present in the DB without carrying out any actual data manipulations. @NoahNsimbe

Baalmart:
Also, just updated the link to the [API documentation of GET Events](https://app.gitbook.com/@airqo/s/airqo-platform-api/device-registry/events#get-events) in the README of this PR. @NoahNsimbe

ja-odur (commented):
I still get the extra fields
I did the following `GET` request `http://127.0.0.1:3000/api/v1/devices/events?tenant=airqo&external=yes`

adding details on how to test the PR will be helpful @Baalmart 

Screenshot 2021-09-29 at 21 19 32

ja-odur (commented):
Ignore the first comment. Everything works fine but maybe for external users, do they need device details? @Baalmart 

Apart from that, everything looks great although I understand @NoahNsimbe has requested some changes.

ja-odur (approved):
LGTM

PR #757, opened 2022-03-21
View microservice build fix

NoahNsimbe:
**_WHAT DOES THIS PR DO?_**

- [x] Removes a test dependency that was causing failure of the workflow. Failing workflow is [here](https://github.com/airqo-platform/AirQo-api/actions/runs/1999004373)
- [x] Adds dependency cache during image building

### DIFF
--- a/src/view/Dockerfile
+++ b/src/view/Dockerfile
@@ -6,6 +6,7 @@ WORKDIR /app
 
 COPY .mvn/ .mvn
 COPY mvnw pom.xml ./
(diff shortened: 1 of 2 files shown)


123MwanjeMike (approved):
LGTM

Baalmart (approved):
LGTM, this microservice needs some marketing so that members can take advantage of it? Members could be missing out, right? @NoahNsimbe

PR #769, opened 2022-03-31
App Forecast Insights

NoahNsimbe:
**_WHAT DOES THIS PR DO?_**

- [x] Creates an ETL pipeline for app insights (Forecast)
- [x] Deletes old insights forecast data.
- [x] Modifies the view api message broker connection, replacing for loops with streams.
- [x] Adds a tasks on the view message broker to delete outdated insights

**_WHAT JIRA STORY/TASK IS RELATED TO THIS PR?_**
[MOB-41](https://airqoteam.atlassian.net/browse/MOB-41)

**_HOW DO I TEST OUT THIS PR?_**
Follow instructions in the [Readme file ](https://github.com/airqo-platform/AirQo-api/tree/app-forecast-etl/src/airflow#running-using-docker)on how to run the app.

**_WHICH ENDPOINTS SHOULD BE READY FOR TESTING?:_** 
App Insights endpoint with the `forecast` query parameter set to `true` 

**EndPoint**: https://platform.airqo.net/api/v1/view/measurements/app/insights?forecast=true
**Documentation**: https://platform.airqo.net/api/v1/view/docs/index.html#App-Insights

PR #776, opened 2022-04-01
move to production

Baalmart:
move to production


NoahNsimbe:
@Baalmart https://github.com/airqo-platform/AirQo-api/pull/769 is ready for production

PR #843, opened 2022-05-15
Fault detection

Paulooh007:
# Fault detection API


Paulooh007:
added fault detection API

PR #883, opened 2022-06-17
move to production

Baalmart:
**_WHAT DOES THIS PR DO?_**
move to production


123MwanjeMike:
I have updated the auth service envs, please confirm that all is well in staging before merge @Baalmart

PR #1002, opened 2022-08-24
making site ID optional

Baalmart:
**_WHAT DOES THIS PR DO?_**
Makes the site ID optional when querying measurements

- Jira cards
    - [PLAT-1201]

```
cd src/device-registry
npm install
```
`npm run dev-mac` or `npm run dev-pc`

add events endpoint

[PLAT-1201]: https://airqoteam.atlassian.net/browse/PLAT-1201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/device-registry/routes/api-v1.js
+++ b/src/device-registry/routes/api-v1.js
@@ -2979,9 +2979,10 @@ router.post(
         .isBoolean()
         .withMessage("is_device_primary should be Boolean"),
       body("*.site_id")
-        .exists()
+        .optional()
+        .notEmpty()
         .trim()
-        .withMessage("site_id is missing")
+        .withMessage("site_id should not be empty if provided")
         .bail()
         .isMongoId()
         .withMessage("site_id must be an object ID")
--- a/src/device-registry/routes/api-v2.js
+++ b/src/device-registry/routes/api-v2.js
@@ -2798,9 +2798,10 @@ router.post(
         .isBoolean()
         .withMessage("is_device_primary should be Boolean"),
       body("*.site_id")
-        .exists()
+        .optional()
+        .notEmpty()
         .trim()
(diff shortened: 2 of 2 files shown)


src/device-registry/routes/api-v1.js
```diff
@@ -2979,9 +2979,10 @@ router.post(
         .isBoolean()
         .withMessage("is_device_primary should be Boolean"),
       body("*.site_id")
-        .exists()
+        .optional()
```
Baalmart, line 2982: in the input validation logic, I have made site_id optional. @NoahNsimbe

src/device-registry/routes/api-v2.js
```diff
@@ -2798,9 +2798,10 @@ router.post(
         .isBoolean()
         .withMessage("is_device_primary should be Boolean"),
       body("*.site_id")
-        .exists()
+        .optional()
```
Baalmart, line 2801: Same case for the V2 endpoint, made SITE_ID optional.

Baalmart (commented):
@NoahNsimbe , I have made the site_id optional in the validation logic.

I also realise other mandatory fields, what should we do for them? They include:

- `device_id` - MONGO_ID
-  `frequency` - STRING
-` time` - DATETIME

Also, we will need to review why we made site_id mandatory in the first place and then find alternative ways of addressing our target goal. Was it about ensuring that we can retrieve measurements by SITE_ID. I guess there was an obsession with SITEID as some point in time.

NoahNsimbe (approved):
LGTM. Thanks @Baalmart

PR #1024, opened 2022-09-01
temporarily disable Kafka in the microservice

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [ ] temporarily disable Kafka in the microservice
- [ ] As this disable is done, it indirectly initiates the update of the ENV of device-registry

### DIFF
--- a/src/device-registry/app.js
+++ b/src/device-registry/app.js
@@ -16,8 +16,8 @@ const { runKafkaConsumer, runKafkaProducer } = require("./config/kafkajs");
 
 mongodb;
 
-runKafkaProducer();
-runKafkaConsumer();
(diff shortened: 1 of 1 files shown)


src/device-registry/app.js
```diff
@@ -16,8 +16,8 @@ const { runKafkaConsumer, runKafkaProducer } = require("./config/kafkajs");
 
 mongodb;
 
-runKafkaProducer();
-runKafkaConsumer();
+// runKafkaProducer();
```
Baalmart, line 19: just disabling the Kafka initialisation in the microservice since it is throwing continuous errors. @NoahNsimbe

PR #1038, opened 2022-09-08
removing the CORS restrictions on the events endpoint

Baalmart:
**_WHAT DOES THIS PR DO?_**
removing the CORS restrictions on the events endpoint. As a hotfix to the netmanager CORS errors when run in local ENV.

### DIFF
--- a/src/device-registry/routes/api-v1.js
+++ b/src/device-registry/routes/api-v1.js
@@ -3131,7 +3131,7 @@ router.post(
 );
 router.get(
   "/events",
-  cors(corsOptionsDelegate),
+  cors(),
   oneOf([
     query("tenant")
       .exists()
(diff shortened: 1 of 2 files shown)


src/device-registry/routes/api-v1.js
```diff
@@ -3131,7 +3131,7 @@ router.post(
 );
 router.get(
   "/events",
-  cors(corsOptionsDelegate),
```
Baalmart, line 3134: updating the CORS configs to allow ALL.

src/device-registry/routes/api-v2.js
```diff
@@ -3061,7 +3061,7 @@ router.post(
 );
 router.get(
   "/events",
-  cors(corsOptionsDelegate),
```
Baalmart, line 3064: updating the CORS configs to allow ALL.

PR #1040, opened 2022-09-08
completely removing the cors middleware

Baalmart:
**_WHAT DOES THIS PR DO?_**
completely removing the cors middleware


src/device-registry/routes/api-v1.js
```diff
@@ -297,7 +297,6 @@ router.get(
 /***list devices */
 router.get(
```
Baalmart, line 298: completely removing the cors middleware from a few GET endpoints for Sites., AirQlouds, and Devices.

PR #1042, opened 2022-09-09
removing cors restrictions

Baalmart:
**_WHAT DOES THIS PR DO?_**
removing cors restrictions


src/device-registry/routes/api-v1.js
```diff
@@ -30,6 +30,17 @@ const corsOptionsDelegate = function(req, callback) {
   callback(null, corsOptions);
 };
 
+const headers = (req, res, next) => {
+  res.header("Access-Control-Allow-Origin", "*");
```
Baalmart, line 34: just allowing all....

Issue #1075, opened 2022-09-21
Enhancements to Explore Data endpoint(s) & email template


FaithDaka:
1. Data endpoint for registering organisations returns a _500_ error. Which endpoint should be used?
![Screenshot (44)](https://user-images.githubusercontent.com/53012890/190729380-092ad856-187f-4046-b382-ddfd698d2080.png)

2. The responding email address should have a user name that defines _AirQo Data_ rather than the name _data_. This is to add professionality to the email responses.
![Screenshot (45)](https://user-images.githubusercontent.com/53012890/190730777-4880e279-f558-4d52-8c4b-7bbecd04778c.jpg)

3. The wording in the email responses requires attention as well. "Once your application is approved, you will receive a confirmation email in this mailbox" reads better.
![Screenshot (46)](https://user-images.githubusercontent.com/53012890/190730154-73b1b468-bb59-4b0a-a9e4-94db1513dbc4.jpg)

123MwanjeMike:
Hi @FaithDaka, Thanks for making these observations.
I have a few comments to share however.
- Firstly, you are getting the response you are getting for not found is 404 under the hood, I think @Baalmart could share the correct endpoint. This is not an issue with the auth-service or the infrastructure.
- For about the professionalism in data@airqo.net vs airqodata@airqo.net, I think the former is better given that the email already has airqo in it, thus indicating the data is from AirQo. The later has repetition.

cc @Baalmart

PR #1085, opened 2022-09-25
addresses issue #1075

Baalmart:
**_WHAT DOES THIS PR DO?_**
- Closes https://github.com/airqo-platform/AirQo-api/issues/1075

- Jira cards
    - [AN-170]
- GitHub issues
    - Closes https://github.com/airqo-platform/AirQo-api/issues/1075

```
cd src/auth-service
npm run stage-mac
```

🚨👇 Please ensure that your `.env` has values for the following variable: 🚨👇

- _MONGO_STAGE_URI_

- [x] [request access](https://airqo-engineering.postman.co/workspace/internal~7a8e9fad-2d13-4d27-af27-f1f4e9a114bf/request/12513372-053764bb-d840-4cf8-88d2-ffbb8a347247)

[AN-170]: https://airqoteam.atlassian.net/browse/AN-170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/auth-service/utils/email.msgs.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
       "we shall get back to you as soon as possible. \n\n" +
       "Before utilising the AirQo data, your application record has to undergo " +
       "the process of approval by the AirQo platform administration.\n" +
-      "Once your application is reviewed, you will receive an email on the mailbox \n\n" +
+      "Once your application is approved, you will receive a confirmation email\n\n" +
```
Baalmart, line 24: The email template message changed as requested.

src/auth-service/services/mailer.js
```diff
@@ -17,6 +17,8 @@ const mailer = {
 
       const mailOptions = {
         from: constants.EMAIL,
+        name: "AirQo Data Team",
```
Baalmart, line 20: adding a sender name to the email in addition to making configurations in Gmail (See attached screenshot)
Unfortunately, this is not bringing the desired change we need., Anyway, let it be in as we continue to investigate. 
Screenshot 2022-09-25 at 18 52 21

src/auth-service/utils/multitenancy.js
```diff
@@ -1,29 +1,27 @@
 const mongodb = require("../config/dbConnection");
 const constants = require("../config/constants");
-const { logElement, logText, logObject } = require("../utils/log");
 
 /****
  * creating a new mongoDB connection by switching tenant
  * using this to create a new connection based on tenant ID
  */
-const getTenantDB = (tenantId, modelName, schema) => {
+function getTenantDB(tenantId, modelName, schema) {
```
Baalmart, line 8: Addressing a potential runtime error relation to TYPES

src/auth-service/utils/request.js
```diff
@@ -106,9 +106,7 @@ const request = {
               data: createdCandidate,
               status,
             });
-          }
-
-          if (responseFromSendEmail.success === false) {
+          } else if (responseFromSendEmail.success === false) {
```
Baalmart, line 109: addressing a potential runtime error and sending HEADERS after they are already sent.

src/auth-service/services/mailer.js
FaithDaka: Thanks for this @Baalmart. 

I think the desired change was to have a professional name for the _data@airqo.net_ email responses. I can see that has been changed. Was there supposed to be another outcome?

Otherwise I see a need to have the **name** attribute in the mail options as a variable, so that it's changed once rather than manually changing it everywhere it appears in the file.

123MwanjeMike (changes requested):
Thanks @Baalmart.
A few comments however.

src/auth-service/config/constants.js
```diff
@@ -16,7 +16,7 @@ const prodConfig = {
 };
 
 const stageConfig = {
-  MONGO_URI: process.env.MONGO_GCE_URI,
+  MONGO_URI: process.env.MONGO_STAGE_URI,
```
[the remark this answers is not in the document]
Baalmart: I already notice this and took care of it by updating the secrets accordingly. @123MwanjeMike 

The challenge with the current approach is that it was hard to differentiate staging and production so it took me a long time to realise that we were using the same variable for staging and production while giving them different values!
The current setup kinda creates a maintenance nightmare.

Happy to hear your additional thoughts.

src/auth-service/services/mailer.js
Baalmart: Thanks @FaithDaka for this. I guess we can definitely make it an ENV. Shall update PR asap.

As for the change, appears it is not "working" after many tests. There is some background magic that could be beyond our control!

Perhaps we just need to do a little more research on this. 🤔

src/auth-service/services/mailer.js
```diff
@@ -17,6 +17,8 @@ const mailer = {
 
       const mailOptions = {
         from: constants.EMAIL,
+        name: constants.EMAIL_NAME,
```
Baalmart, line 20: @FaithDaka 
The name is now a constant variable. I couldn't use ENV because of the spaces in the name.

Otherwise, please note that the emails still come in as "data" despite all our efforts! 😁😁

Happy to hear additional thoughts/feedback.

Thanks!

src/auth-service/config/constants.js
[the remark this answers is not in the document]
123MwanjeMike: Hi @Baalmart 
The concern I have with it is the redundancy the approach creates.
It also means we are having production secrets added to the staging .env which is security wise not advisable.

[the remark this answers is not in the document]
Baalmart: @123MwanjeMike 
Perhaps we should not have the production one added to the staging ENV. 
Perhaps you have just not realised the maintenance nightmare this current approach is.

Imagine, as a developer, having to update the value of the ENV variable when testing out different environment. Instead of just running different commands.

It makes development more cumbersome and yet the shared security concerns are some debatable.

Happy to jump on a call to showcase the additional complexities the current approach brings.

In summary, what the current approach attempts to do is already taken care of in the config file of the microservice(s). The ENVS are supposed to be different variables for reasons already shared above.

src/auth-service/services/mailer.js
Baalmart: Hi @FaithDaka , this has now been modified as per the learnings from another PR: https://github.com/airqo-platform/AirQo-api/pull/1119

You can now test out and share more feedback accordingly.

Thanks!

src/auth-service/config/constants.js
```diff
@@ -1,13 +1,13 @@
...
+  MONGO_URI: process.env.MONGO_DEV_URI,
   DB_NAME: process.env.MONGO_DEV,
   PWD_RESET: `${process.env.PLATFORM_DEV_BASE_URL}/reset`,
   LOGIN_PAGE: `${process.env.PLATFORM_DEV_BASE_URL}/login`,
   FORGOT_PAGE: `${process.env.PLATFORM_DEV_BASE_URL}/forgot`,
   PLATFORM_BASE_URL: process.env.PLATFORM_DEV_BASE_URL,
 };
 const prodConfig = {
-  MONGO_URI: process.env.MONGO_GCE_URI,
+  MONGO_URI: process.env.MONGO_PROD_URI,
```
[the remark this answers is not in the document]
Baalmart: Hi @123MwanjeMike , all the logic for having one variable is handled inside the configuration files of the microservices.

The ones in the ENV files are supposed to be named differently.

Perhaps we need to have a proper onboarding for some of these microservices as we continue to refactor them.

Thanks!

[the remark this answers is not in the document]
Baalmart: **Additional notes:**
In the future, whenever we make these changes, we will want to take them through a PR review process. Just like what we are doing here in this PR. We should not discover such changes by chance. @123MwanjeMike

123MwanjeMike (approved):

src/auth-service/config/constants.js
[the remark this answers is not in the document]
123MwanjeMike: Thanks @Baalmart 
I think I now understand the exhaustion this creates during development.

Although, what I was trying to break is the dependency of the codebase to the environment. In that we(would) always have update the code whenever we have a new environment for the api to run in. In other words we would have 5 different environment variables for just the database connection if we had 5 separate environments instead of just the 2 (stage and prod) we currently have.

But we can for now proceed with it and refactor it in the future.

PR #1140, opened 2022-10-21
introducing new script for running staging in device-registry

Baalmart:
**_WHAT DOES THIS PR DO?_**
introducing new script for running staging in device-registry
We need to ensure that the appropriate changed files function as expected after deployment

```
cd src/device-registry
npm install
npm run stage
```


123MwanjeMike (approved):
LGTM

PR #1229, opened 2022-12-15
only calculate uptime for AirQo devices

Baalmart:
**_WHAT DOES THIS PR DO?_**
- only calculate uptime for AirQo devices


src/device-uptime/device_uptime.py
```diff
@@ -43,6 +44,8 @@ def save_device_uptime(tenant):
     active_device_count = 0
 
     for device in devices:
+        if device.get("network", "") != "airqo":
```
Baalmart, line 47: Only consider AirQo devices when calculating Uptime. cc: @sserurich

PR #1362, opened 2023-02-23
log error to Slack in case message sent is undefined

Baalmart:
**_WHAT DOES THIS PR DO?_**
log error to Slack in case message sent is undefined


src/device-registry/bin/www
```diff
@@ -60,6 +60,13 @@ const run = async() => {
     await consumer.subscribe({ topic: constants.HOURLY_MEASUREMENTS_TOPIC, fromBeginning: true });
     await consumer.run({
         eachMessage: async({ topic, partition, message }) => {
+            if (isEmpty(message)) {
+                logger.error(
+                    `KAFKA: the sent message in undefined --- ${JSON.stringify(
```
Baalmart, line 65: log the error in case the message is undefined. @NoahNsimbe

PR #1365, opened 2023-02-28
removing some input validations from the device creation process

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [ ] removing some input validations from the device creation process

### DIFF
--- a/src/device-registry/routes/v1/devices.js
+++ b/src/device-registry/routes/v1/devices.js
@@ -443,10 +443,7 @@ router.post(
       )
       .bail()
       .notEmpty()
-      .withMessage("the name should not be empty if provided")
-      .bail()
-      .matches(/^[a-zA-Z0-9]+$/)
-      .withMessage("name should only contain alphanumeric characters."),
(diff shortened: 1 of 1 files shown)


src/device-registry/routes/v1/devices.js
```diff
@@ -443,21 +443,15 @@ router.post(
       )
       .bail()
       .notEmpty()
-      .withMessage("the name should not be empty if provided")
-      .bail()
-      .matches(/^[a-zA-Z0-9]+$/)
```
Baalmart, line 448: no need to do input validations for` long_name` and `name` during device creation

src/device-registry/routes/v1/devices.js
```diff
@@ -1112,21 +1106,15 @@ router.post(
...
```
Baalmart, line 1117: no need to do input validations for `long_name`  and `name` during soft device creation

Codebmk (commented):
LGTM Thanks
Screenshot from 2023-02-28 14-59-30

PR #1420, opened 2023-03-27
[HEATMAP] Refactored: Data and Data preprocessing

uman95:
**_WHAT DOES THIS PR DO?_**

- Uses site data (as supposed to device data)
- Proper data preprocessing
- Remove spatial-temporal modelling for country-level AirQlouds

- Jira cards
    - [Refined spatial models for one international AirQloud](https://airqoteam.atlassian.net/browse/DP-205)
- GitHub issues
    - Closes #<enter GitHub issue number here>

- create a virtual environment
- install all requirements, src/gpmodel/requirements.txt
- get the .env and test using development environment (**note that: for the event API we now use .../api/v2/..**)
- run the src/gpmodel/main.py

The [**heatmap** API](https://staging-platform.airqo.net/api/v1/predict/heatmap)

### DIFF
--- a/src/gp-model/config.py
+++ b/src/gp-model/config.py
@@ -23,6 +23,7 @@ class ProductionConfig(Config):
     VIEW_AIRQLOUD_URI = os.getenv('VIEW_AIRQLOUD_URI_PROD')
     LIST_DEVICES_URI = os.getenv('LIST_DEVICES_URI_PROD')
     EVENTS_URI = os.getenv('EVENTS_URI_PROD')
+    API_TOKEN = os.getenv('PROD_TOKEN')
 
 
 class DevelopmentConfig(Config):
@@ -35,6 +36,7 @@ class DevelopmentConfig(Config):
     VIEW_AIRQLOUD_URI = os.getenv('VIEW_AIRQLOUD_URI_DEV')
     LIST_DEVICES_URI = os.getenv('LIST_DEVICES_URI_DEV')
     EVENTS_URI = os.getenv('EVENTS_URI_DEV')
+    API_TOKEN = os.getenv('DEV_TOKEN')
 
 
 class TestingConfig(Config):
--- a/src/gp-model/config.py
+++ b/src/gp-model/config.py
@@ -47,6 +47,7 @@ class TestingConfig(Config):
     VIEW_AIRQLOUD_URI = os.getenv('VIEW_AIRQLOUD_URI_STAGE')
     LIST_DEVICES_URI = os.getenv('LIST_DEVICES_URI_STAGE')
     EVENTS_URI = os.getenv('EVENTS_URI_STAGE')
+    API_TOKEN = os.getenv('STAGE_TOKEN')
 
 
 app_config = {
--- a/src/gp-model/data/data.py
+++ b/src/gp-model/data/data.py
@@ -1,0 +1,76 @@
+import requests
+from datetime import datetime, timedelta
+from pymongo import MongoClient
+from config import configuration, environment
+
+
+def get_airqloud_sites(airqloud, tenant='airqo'):
+
+    # airqloud_url = 'https://platform.airqo.net/api/v1/devices/airqlouds'
+    params = {'tenant': tenant,
+              'airqloud': airqloud
+
+              }
+
+    if environment == 'staging':
+        headers = {'Authorization': configuration.API_TOKEN}
+        airqloud_response = requests.get(
+            configuration.VIEW_AIRQLOUD_URI, params=params, headers=headers)
+    else:
+        airqloud_response = requests.get(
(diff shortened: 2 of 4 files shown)


src/gp-model/data/data.py
```diff
@@ -0,0 +1,76 @@
...
+    site_ids = [site['_id'] for site in sites_details]
+
+    return site_ids
+
+
+def get_site_data(site_id, tenant='airqo'):
+    start_time = (datetime.utcnow()-timedelta(days=7)
+                  ).strftime('%Y-%m-%dT%H:%M:%SZ'),
+    end_time = datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ')
+    event_url = 'https://platform.airqo.net/api/v2/devices/events'
```
NoahNsimbe, line 35: Hi @uman95, we can use an environment variable instead of hard coding `https://platform.airqo.net`

uman95: Thanks @NoahNsimbe it was an oversight, thought I've cleaned it all up. New commit now available.

NoahNsimbe (approved):
Thanks @uman95. The code changes LGTM. Let a suggestion(s) for your consideration.

src/gp-model/main.py
```diff
@@ -234,46 +163,44 @@ def periodic_function(tenant, airqloud, aq_id):
...
+        airqlouds = requests.get(
+            configuration.VIEW_AIRQLOUD_URI, params=params).json()['airqlouds']
     names = [aq['name'] for aq in airqlouds]
     aq_ids = [aq['_id'] for aq in airqlouds]
-    # Esxcluding Uganda
-    del names[-1:]
-    del aq_ids[-1:]
-    return names, aq_ids
+    # Esxcluding Country level AirQloud
+    exclude = ['kenya', 'uganda', 'cameroon', 'senegal', 'gulu']
```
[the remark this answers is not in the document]
NoahNsimbe: Screenshot 2023-03-27 at 22 13 42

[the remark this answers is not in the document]
uman95: Thanks @NoahNsimbe will address it.

src/gp-model/config.py
```diff
@@ -47,6 +47,7 @@ class TestingConfig(Config):
     VIEW_AIRQLOUD_URI = os.getenv('VIEW_AIRQLOUD_URI_STAGE')
     LIST_DEVICES_URI = os.getenv('LIST_DEVICES_URI_STAGE')
     EVENTS_URI = os.getenv('EVENTS_URI_STAGE')
+    API_TOKEN = os.getenv('STAGE_TOKEN')
```
123MwanjeMike, line 50: Could we have this API_TOKEN environment independent?

src/gp-model/main.py
```diff
@@ -55,92 +22,54 @@ def get_airqloud_polygon(tenant, airqloud):
     params = {'tenant': tenant,
               'name': airqloud
               }
-    coords = requests.get(configuration.VIEW_AIRQLOUD_URI, params=params).json()[
-        'airqlouds'][0]['location']['coordinates']
+    if environment == 'staging':
+        headers = {'Authorization': configuration.API_TOKEN}
```
123MwanjeMike, line 26: We also have a new authentication method that use Access token passed as a query parameter which I find to be a cleaner approach.

123MwanjeMike (changes requested):
Great work @uman95
I just a few comments on this.
They may not be a big concern but will leave them here so we don't miss them.
Thanks

sserurich (commented):
@uman95 , running PR locally with staging environment fails with this message in the image, might need to catch error incase response from the request doesn't have 'airqloud'
image

uman95:
> @uman95 , running PR locally with staging environment fails with this message in the image, might need to catch error incase response from the request doesn't have 'airqloud' image

Hi @sserurich you need an API token now to test out the code in a staging environment. Also check if you are using **version 2 (v2)** in all your APIs because we are moving to version 2 now.

sserurich:
> @uman95 , running PR locally with staging environment fails with this message in the image, might need to catch error incase response from the request doesn't have 'airqloud' image

The issue stems from token not working.. but in the future, might need to catch this as well.

sserurich (approved):
LGTM, now

src/gp-model/main.py
uman95: @123MwanjeMike done.

src/gp-model/config.py
uman95: Thanks @123MwanjeMike I have made it independent.

123MwanjeMike (approved):
Great work @uman95 
LGTM

src/gp-model/main.py
123MwanjeMike: Great, thanks @uman95

PR #1441, opened 2023-04-01
disable Slack logging for duplicate data entries

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [x] disable Slack logging for duplicate data entries

### DIFF
--- a/src/device-registry/bin/www
+++ b/src/device-registry/bin/www
@@ -147,11 +147,11 @@ const run = async() => {
                         logObject("responseFromInsertMeasurements", responseFromInsertMeasurements)
 
                         if (responseFromInsertMeasurements.success === false) {
-                            logger.error(
-                                `KAFKA: responseFromInsertMeasurements --- ${JSON.stringify(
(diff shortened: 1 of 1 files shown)


src/device-registry/bin/www
```diff
@@ -147,11 +147,11 @@ const run = async() => {
                         logObject("responseFromInsertMeasurements", responseFromInsertMeasurements)
 
                         if (responseFromInsertMeasurements.success === false) {
-                            logger.error(
-                                `KAFKA: responseFromInsertMeasurements --- ${JSON.stringify(
-                      responseFromInsertMeasurements
-                    )}`
-                            );
+                            //         logger.error(
```
Baalmart, line 150: I am just commenting out the Slack logging stuff so that our Slack channel is LESS noisy (spammy) @BenjaminSsempala

Baalmart: this is the part where the consumed measurements from KAFKA are NOT successfully inserted in device registry. @BenjaminSsempala

PR #1481, opened 2023-04-19
ability to query KCCA devices individually

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [x] ability to query KCCA devices individually

### DIFF
--- a/src/device-registry/utils/generate-filter.js
+++ b/src/device-registry/utils/generate-filter.js
@@ -314,7 +314,7 @@ const generateFilter = {
   },
   devices: (req) => {
     try {
-      let filter = {};
+      let filter = { name: {} };
       let {
         name,
         channel,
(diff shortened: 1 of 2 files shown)


src/device-registry/utils/generate-filter.js
```diff
@@ -340,24 +340,46 @@ const generateFilter = {
...
-        filter["name"] = device;
-      }
-
-      if (device_name) {
-        // let regexExpression = generateFilter.generateRegexExpressionFromStringElement(
-        //   name
-        // );
-        filter["name"] = device_name;
+        let deviceArray = name.split(",");
+        let modifiedDeviceArray = deviceArray.map((value) => {
```
Baalmart, line 344: whether provided as lowercase or uppercase, query should be possible.

PR #1485, opened 2023-04-20
ensure any device's long name is also unique

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [x] ensure any device's long name is also unique


src/device-registry/models/Device.js
```diff
@@ -71,6 +71,7 @@ const deviceSchema = new mongoose.Schema(
     long_name: {
       type: String,
       trim: true,
+      unique: true,
```
Baalmart, line 74: Just to prevent metadata issues from coming up in the system.

PR #1610, opened 2023-06-01
quick fix to import issue

uman95:
**_WHAT DOES THIS PR DO?_**

Fix import issue. Only added the code below:

`from threading import Thread`


Baalmart (approved):
thanks @uman95

PR #1741, opened 2023-06-25
Implementation of hourly forecasts

Mnoble-19:
**_WHAT DOES THIS PR DO?_**
* updates the training and prediction job for the hourly forecasts model
* Fixes issue of missing forecasts for some sites
* Code fixes for daily forecast job as well
* Add saving data to BigQuery
* Remove redudant code

hourly & daily forecast endpoints


src/predict/jobs/forecast/Dockerfile
```diff
@@ -14,13 +14,16 @@ RUN pip install -r requirements.txt
 FROM base as production
 ENV ENV=production
 CMD ["python", "daily_forecasts.py"]
+CMD ["python", "hourly_forecasts.py"]
 
 # Staging
 FROM base as staging
 ENV ENV=staging
 CMD ["python", "daily_forecasts.py"]
+CMD ["python", "hourly_forecasts.py"]
```
Mnoble-19, line 23: Hi @NoahNsimbe and @123MwanjeMike. I need your help here. These 2 scripts are separate jobs and ought to run at different times. Is this okay? How does one discern between them which one to run at a given period

NoahNsimbe: @123MwanjeMike

123MwanjeMike: Hi @Mnoble-19 
Why are you having two different entry points(`CMD`) in the same stage dockerfile.
Can we use different stages for this.
https://docs.docker.com/build/building/multi-stage/

123MwanjeMike (changes requested):

src/predict/jobs/forecast/Dockerfile
```diff
@@ -11,16 +11,29 @@ RUN pip install --upgrade pip
 RUN pip install -r requirements.txt
 
 # Production
```
123MwanjeMike, line 13: Thanks @Mnoble-19 
Could we also have these comments updated to something like “production daily” and “production hourly”

123MwanjeMike (approved):

NoahNsimbe (approved):
LGTM

PR #1791, opened 2023-07-02
Setup automated tweets

Mnoble-19:
**_WHAT DOES THIS PR DO?_**
* Sets up automated tweets for `AirQo Engineering` account


Mnoble-19:
@NoahNsimbe and @123MwanjeMike would like some insights from you guys on how to deploy this. It's essentially a job meant to run a specific time, but IMO k8s may be overkill for it as it doesn't need significant resources, I'm thinking of  having it as a `cloud function`. Share thoughts.

123MwanjeMike:
> @NoahNsimbe and @123MwanjeMike would like some insights from you guys on how to deploy this. It's essentially a job meant to run a specific time, but IMO k8s may be overkill for it as it doesn't need significant resources, I'm thinking of having it as a `cloud function`. Share thoughts.

It's better to deploy it on K8s for easier management along with the other micros. We can give it the least resources from the cluster so that its consumption is kept in check.

123MwanjeMike (changes requested):
Great work @Mnoble-19 
I however notice the folder structure is tied to a specific cloud deployment tool(cloud functions). Could we remove this dependency even at the folder structure?

PR #1811, opened 2023-07-05
KAFKA errors: only timestamp and device-number as the only logged 

Baalmart:
**_WHAT DOES THIS PR DO?_**
timestamp and device-number as the only logged detailed outside the error values

### DIFF
--- a/src/device-registry/bin/www
+++ b/src/device-registry/bin/www
@@ -150,17 +150,8 @@ const run = async() => {
                                 message: detail.message,
                                 key: detail.context.key,
                                 value: detail.context.value,
-                                path: detail.path,
                                 device_number: device_number ? device_number : undefined,
-                                device_id: device_id ? device_id : undefined,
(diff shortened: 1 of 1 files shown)


src/device-registry/bin/www
```diff
@@ -150,17 +150,8 @@ const run = async() => {
                                 message: detail.message,
                                 key: detail.context.key,
                                 value: detail.context.value,
-                                path: detail.path,
                                 device_number: device_number ? device_number : undefined,
-                                device_id: device_id ? device_id : undefined,
                                 timestamp: timestamp ? timestamp : undefined,
-                                site_id: site_id ? site_id : undefined,
-                                battery: battery ? battery : undefined,
-                                network: network ? network : undefined,
```
Baalmart, line 159: just to removed other details just to ensure that we do not clutter Slack notifs with uneeded details @BenjaminSsempala

PR #1824, opened 2023-07-09
fixing a runtime error for mongoose projection

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [ ] fixing a runtime error for mongoose projection

### DIFF
--- a/src/device-registry/models/Device.js
+++ b/src/device-registry/models/Device.js
@@ -462,7 +462,14 @@ deviceSchema.statics = {
         .allowDiskUse(true);
 
       if (Object.keys(exclusionProjection).length > 0) {
-        pipeline.project(exclusionProjection);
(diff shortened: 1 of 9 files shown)


src/device-registry/models/Device.js
```diff
@@ -462,7 +462,14 @@ deviceSchema.statics = {
         .allowDiskUse(true);
 
       if (Object.keys(exclusionProjection).length > 0) {
-        pipeline.project(exclusionProjection);
+        // Construct the $project stage dynamically with the exclusionProjection
```
Baalmart, line 465: Hi @BenjaminSsempala , please take note of this runtime error that has just been fixed and apply the same fix in `auth-service` in case it is also needed there? Thanks!
Reference can be made to the numerous Slack notifications inside `#notifs-device-registry`

PR #1831, opened 2023-07-10
undoing previous Model projection changes when listing

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [x] undoing previous Model projection changes when listing

### DIFF
--- a/src/auth-service/config/constants.js
+++ b/src/auth-service/config/constants.js
@@ -531,11 +531,11 @@ const defaultConfig = {
     longitude: 1,
     reference_site: 1,
     place_id: 1,
-    firebase_user_id:1,
+    firebase_user_id: 1,
   },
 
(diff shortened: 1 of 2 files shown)


src/auth-service/config/constants.js
```diff
@@ -531,11 +531,11 @@ const defaultConfig = {
     longitude: 1,
     reference_site: 1,
     place_id: 1,
-    firebase_user_id:1,
+    firebase_user_id: 1,
   },
 
   FAVORITES_EXCLUSION_PROJECTION: (category) => {
-    const initialProjection = {  };
+    const initialProjection = { nothing: 0 };
```
Baalmart, line 538: @BenjaminSsempala , this has turned out to be the best approach going forward to similar situations. The previous approach was buggy. :)

PR #1833, opened 2023-07-11
Update dockerfiles for forecast jobs

Mnoble-19:
**_WHAT DOES THIS PR DO?_**
- Combines forecast job & training stages into one.

### DIFF
--- a/src/predict/jobs/forecast/Dockerfile
+++ b/src/predict/jobs/forecast/Dockerfile
@@ -10,30 +10,18 @@ WORKDIR /app/
 RUN pip install --upgrade pip
 RUN pip install -r requirements.txt
 
-# Production - daily forecasts
-FROM base as production_daily
+# Production
+FROM base as production
 ENV ENV=production
-CMD ["python", "daily_forecasts.py"]
+CMD ["sh", "run_scripts.sh"]
 
-# Production - hourly forecasts
-FROM base as production_hourly
-ENV ENV=production
-CMD ["python", "hourly_forecasts.py"]
 
-# Staging - daily forecasts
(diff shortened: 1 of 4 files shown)


src/predict/jobs/forecast/Dockerfile
```diff
@@ -10,30 +10,18 @@ WORKDIR /app/
 RUN pip install --upgrade pip
 RUN pip install -r requirements.txt
 
-# Production - daily forecasts
-FROM base as production_daily
+# Production
+FROM base as production
 ENV ENV=production
-CMD ["python", "daily_forecasts.py"]
+CMD ["sh", "run_scripts.sh"]
```
Mnoble-19, line 16: @123MwanjeMike . I tried to have 2 CMDs for each script but got the warning. `There can only be one CMD instruction in a Dockerfile or build stage. Only the last one will have an effect` so I've just used a bash script. Don't know why I didn't consider that first. 😅🤡🤦🏾‍♂️

123MwanjeMike: Thanks for this improvement @Mnoble-19 
For the 3-5minutes timespan, the VMs run for 8minutes by default so we are good.

123MwanjeMike (approved):
This is a great improvement.
Thanks @Mnoble-19

123MwanjeMike (changes requested):
Hi @Mnoble-19 
I'm seeing the error log below failing the deploy preview. Could we have it fixed too?

image

123MwanjeMike (approved):

NoahNsimbe (approved):
LGTM

PR #1890, opened 2023-07-21
fix the gateway timeout

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [ ] fix the gateway timeout when querying events

### DIFF
--- a/src/device-registry/utils/create-event.js
+++ b/src/device-registry/utils/create-event.js
@@ -13,7 +13,6 @@ const { transform } = require("node-json-transform");
 const Dot = require("dot-object");
 const cleanDeep = require("clean-deep");
 const { getDevicesCount, list, decryptKey } = require("./create-monitor");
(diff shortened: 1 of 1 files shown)


src/device-registry/utils/create-event.js
```diff
@@ -1089,29 +1090,33 @@ const createEvent = {
...
             message: "no cache present",
             errors: { message: "no cache present" },
+            status: httpStatus.INTERNAL_SERVER_ERROR,
           });
         }
       });
     } catch (error) {
       logger.error(`internal server error -- ${error.message}`);
-      return {
+      callback({
```
Baalmart, line 1114: @BenjaminSsempala , this was causing the timeout issue. The function utilises a callback function as an input and yet we were just returning fwwaaaa, like other normal parameter funcs....  cc: @NoahNsimbe

PR #1950, opened 2023-08-02
move to production

Baalmart:
- [ ] https://github.com/airqo-platform/AirQo-api/pull/1949


src/device-registry/utils/create-know-your-air.js
```diff
@@ -455,10 +455,13 @@ const createKnowYourAir = {
 
           if (responseFromListProgress.data.length == 0) {
```
Baalmart, line 456: @BenjaminSsempala , in future PRs, use the tripple (===) instead of the double (==). You can read more about the reason behind this on the internet.

BenjaminSsempala: Thank you @Baalmart I'll update this in any followup PRs

PR #1988, opened 2023-08-08
Fix heatmap endpoint

Mnoble-19:
Fixes the heatmap endpoint to return all values

### DIFF
--- a/src/predict/api/controllers/prediction.py
+++ b/src/predict/api/controllers/prediction.py
@@ -1,26 +1,20 @@
 import logging
 import traceback
-from app import cache
 
 from dotenv import load_dotenv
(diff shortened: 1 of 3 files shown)


src/gp-model/main.py
```diff
@@ -201,7 +201,7 @@ def predict_model(m, tenant, airqloud, aq_id, poly, x1, x2, y1, y2):
     result.append(result_builder)
 
     db = connect_mongo(tenant)
-    collection = db["gp_predictions"]
+    collection = db["gp_model_predictions"]
```
Mnoble-19, line 204: Hi @uman95 , the old collection has some old "dirty" data and I don't have access to delete it, so Martin suggested I just create a new one

PR #2065, opened 2023-08-16
Faulty devices hotfix

Mnoble-19:
**_WHAT DOES THIS PR DO?_**
-Simplify process of capturing query parameters
-Currently the endpoint returns   an empty array because the token is included in one of the parameters for querying the DB
Screenshot 2023-08-16 at 19 45 57


NoahNsimbe (approved):
LGTM

PR #2086, opened 2023-08-19
re-enable rate limiting

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [x] re-enable rate limiting

### DIFF
--- a/src/auth-service/routes/v1/tokens.js
+++ b/src/auth-service/routes/v1/tokens.js
@@ -236,7 +236,7 @@ router.get(
         .withMessage("the token must not be empty"),
     ],
   ]),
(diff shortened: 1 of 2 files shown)


src/auth-service/routes/v1/tokens.js
```diff
@@ -236,7 +236,7 @@ router.get(
         .withMessage("the token must not be empty"),
     ],
   ]),
-  // rateLimitMiddleware,
+  rateLimitMiddleware,
```
Baalmart, line 239: @BenjaminSsempala , this is the middleware for re-enabling rate limiting....

PR #2104, opened 2023-08-22
refactoring the models to export models instead of schemas

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [x] refactoring the models to export models instead of schemas
- [x] fix a runtime error around the verify JWT user endpoint

### DIFF
--- a/src/auth-service/controllers/create-user.js
+++ b/src/auth-service/controllers/create-user.js
@@ -223,41 +223,9 @@ const createUser = {
     logText("..................................");
     logText("user verify......");
     try {
-      const hasErrors = !validationResult(req).isEmpty();
-      if (hasErrors) {
-        let nestedErrors = validationResult(req).errors[0].nestedErrors;
-        logger.error(
-          `input validation errors ${JSON.stringify(
-            convertErrorArrayToObject(nestedErrors)
-          )}`
-        );
-        return badRequest(
-          res,
-          "bad request errors",
-          convertErrorArrayToObject(nestedErrors)
-        );
-      }
-
-      logObject("req.user", req.user);
-      logObject("req.user.toAuthJSON()", req.user.toAuthJSON());
-
-      if (req.auth.success === true) {
-        return res.status(httpStatus.OK).send("this token is valid");
-      } else {
-        if (req.auth.error) {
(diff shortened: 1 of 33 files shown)


src/auth-service/models/Client.js
```diff
@@ -227,4 +228,14 @@ ClientSchema.methods = {
   },
 };
 
-module.exports = ClientSchema;
+const ClientModel = (tenant) => {
```
Baalmart, line 231: @BenjaminSsempala , now all models export the model instead of the Schema.

src/auth-service/utils/control-access.js
```diff
@@ -34,96 +33,6 @@ const generateClientSecret = (length) => {
...
-  try {
-    let tokens = mongoose.model("access_tokens");
-    return tokens;
-  } catch (error) {
-    let tokens = getModelByTenant(tenant, "access_token", AccessTokenSchema);
-    return tokens;
-  }
-};
-
-const PermissionModel = (tenant) => {
```
Baalmart, line 67: @BenjaminSsempala , because of this refactor, then we just import the models directly and pass in the tenant parameter. No need for such now.

src/auth-service/controllers/create-user.js
```diff
@@ -220,7 +220,18 @@ const createUser = {
   },
 
   verify: (req, res) => {
-    return res.status(httpStatus.OK).send("this token is valid");
+    logText("..................................");
+    logText("user verify......");
```
Baalmart, line 224: @BenjaminSsempala , the other error we are getting via Slack is here but it is a bit hard to replicate or even understand fully! :) 

So the purpose of this PR just changed to refactoring since I could not fix this.

Baalmart (commented):
@BenjaminSsempala , please review the code acccordingly

src/auth-service/utils/control-access.js
BenjaminSsempala: Thanks @Baalmart

src/auth-service/models/Client.js
BenjaminSsempala: Thanks @Baalmart

PR #2122, opened 2023-08-24
move to production

Baalmart:
- [ ] https://github.com/airqo-platform/AirQo-api/pull/2121

### DIFF
--- a/k8s/analytics/values-prod.yaml
+++ b/k8s/analytics/values-prod.yaml
@@ -8,7 +8,7 @@ images:
     celeryWorker: eu.gcr.io/airqo-250220/airqo-analytics-celery-worker
     reportJob: eu.gcr.io/airqo-250220/airqo-analytics-report-job
     devicesSummaryJob: eu.gcr.io/airqo-250220/airqo-analytics-devices-summary-job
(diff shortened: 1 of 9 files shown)


src/device-registry/routes/v2/kya.js
```diff
@@ -1902,10 +1907,20 @@ router.put(
...
       body("content")
-        .optional()
-        .notEmpty()
-        .withMessage("the content should not be empty")
-        .trim(),
+        .isArray()
+        .withMessage("content should be an array")
+        .bail()
+        .custom((value) => {
+          if (!Array.isArray(value)) {
```
Baalmart, line 1914: @BenjaminSsempala ...the repetition is still present but we shall check it out in a future PR I guess

BenjaminSsempala: That's odd.
Sure thing I'll remember to remove it in a future pr.
@Baalmart

BenjaminSsempala: Ooooh I didn't remove it in the validation for update,
Only the one for create, I'll handle this in a different PR. Thanks @Baalmart

PR #2151, opened 2023-09-04
Faulty devices endpoint update

Mnoble-19:
**_WHAT DOES THIS PR DO?_**
* Updates `faulty-devices` job to use `upsert` rather than `inserting` new records

* The endpoint returns more data than is required, and some results are duplicated


src/airflow/airqo_etl_utils/airqo_utils.py
```diff
@@ -968,5 +968,6 @@ def save_faulty_devices(data: pd.DataFrame):
         client = pm.MongoClient(configuration.MONGO_URI)
         db = client[configuration.MONGO_DATABASE_NAME]
         records = data.to_dict("records")
+        db.faulty_devices.delete_many({})
```
Baalmart, line 971: @Mnoble-19 , could you shed more light on this? This looks to be an "expensive" operation. On which DB are we storing these records?

Baalmart (changes requested):
Hi @Mnoble-19 
Operations of deleting many records can be avoided in so many ways so disapproving this PR. cc: @BenjaminSsempala

src/airflow/airqo_etl_utils/airqo_utils.py
```diff
@@ -964,9 +967,28 @@ def flag_faults(df):
...
+                for record in records
+            ]
+
+            try:
+                db.faulty_devices.bulk_write(bulk_ops)
+            except Exception as e:
+                print(f"Error saving faulty devices to MongoDB: {e}")
+
+            current_device_names = [record["device_name"] for record in records]
+            db.faulty_devices.delete_many(
```
Baalmart, line 990: Hi @Mnoble-19 , please shed more light on this delete many operation?
If possible, please use PR review comments on all the changed files to make our review process easier?
Thanks!

Baalmart (approved):

PR #2159, opened 2023-09-06
Translations to French for health tips from the API

BenjaminSsempala:
**_WHAT DOES THIS PR DO?_**
- Translations to French for health tips from the API

- Jira cards
    - []()
- GitHub issues
    - Closes #<enter GitHub issue number here>

### DIFF
--- a/src/device-registry/utils/create-health-tips.js
+++ b/src/device-registry/utils/create-health-tips.js
@@ -32,7 +32,7 @@ const createHealthTips = {
         limit,
         skip,
       });
-      if (language != null) {
+      if (language !== null) {
         responseFromListHealthTips.data = await translateUtil.translate(responseFromListHealthTips.data, language);
       }
 
--- a/src/device-registry/utils/create-health-tips.js
+++ b/src/device-registry/utils/create-health-tips.js
@@ -26,14 +26,16 @@ const createHealthTips = {
       const skip = parseInt(request.query.skip, 0);
       const filter = generateFilter.tips(request);
       const language = request.query.language;
+      let translatedHealthTips;
 
-      const responseFromListHealthTips = await HealthTipModel(tenant).list({
+      let responseFromListHealthTips = await HealthTipModel(tenant).list({
         filter,
         limit,
         skip,
       });
-      if (language !== null) {
-        responseFromListHealthTips.data = await translateUtil.translate(responseFromListHealthTips.data, language);
+      if (language) {
+        translatedHealthTips = await translateUtil.translate(responseFromListHealthTips.data, language);
+        responseFromListHealthTips = translatedHealthTips;
       }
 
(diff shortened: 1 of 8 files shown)


src/device-registry/utils/create-health-tips.js
```diff
@@ -24,12 +25,17 @@ const createHealthTips = {
...
       const skip = parseInt(request.query.skip, 0);
       const filter = generateFilter.tips(request);
+      const language = request.query.language;
 
       const responseFromListHealthTips = await HealthTipModel(tenant).list({
         filter,
         limit,
         skip,
       });
+      if (language != null) {
```
Baalmart, line 35: Hi @BenjaminSsempala 
Please consider using `!==` comparison operator for benefits you can search up online. Thanks!

src/device-registry/utils/translate.js
```diff
@@ -0,0 +1,50 @@
+const constants = require("@config/constants");
+const log4js = require("log4js");
+const logger = log4js.getLogger(
```
Baalmart, line 3: Hi @BenjaminSsempala ....in an effort to minimise the number of "new files", could you add these as "helper functions" above the `create-health-tips `module?
WHY? I am just worried of an exponential number of new files which can consequentially make our microservice intimidating to a new user. Additionally, it can also make maintenance tricky. One major benefit of the concept of "reusing" is that it forces us to appreciate what is already existing in our codebases instead of just piling new files for every new functionality.

 Suggested refactoring above is more like below: 

const translate = (healthTips, targetLanguage) => { 
// function body
}
const translateText = (text, target) => {
// function body
}


const createHealthTips = {
  list: async (request) => { },
  fn2: async (request) => {},
 ......
}
module.exports = createHealthTips

BenjaminSsempala: Thanks @Baalmart 
I'd added this as a separate util file because the same util will also be utilised by the KYA lessons and quiz or any other API output that needs translation from the mobile app.
Would you recommend that each Util has its own translate helper function?

src/device-registry/utils/create-health-tips.js
BenjaminSsempala: Thanks @Baalmart

src/device-registry/utils/translate.js
Baalmart: Okay, thanks for the clarification @BenjaminSsempala !

Baalmart: I think we can maintain the current structure. Explanation makes sense, as long as we do not create new file for every new function, we shall keep the file size manageable. @BenjaminSsempala

Baalmart (approved):
Thanks @BenjaminSsempala , all is good now. Thanks!

PR #2178, opened 2023-09-13
re-enable user update emails

Baalmart:
- [x] re-enable user update emails


src/auth-service/utils/create-user.js
```diff
@@ -247,12 +247,7 @@ const createUserModule = {
       if (responseFromModifyUser.success === true) {
         const { _id, ...updatedUserDetails } = responseFromModifyUser.data;
         logObject("updatedUserDetails", updatedUserDetails);
-        return {
-          success: true,
```
Baalmart, line 251: Due to some recent metadata updates for users, we had temporarily disabled the emails which are sent out whenever a user updates their details. @BenjaminSsempala / @Codebmk

PR #2181, opened 2023-09-13
move to production

Baalmart:
- [ ] https://github.com/airqo-platform/AirQo-api/pull/2180
- [ ] https://github.com/airqo-platform/AirQo-api/pull/2159
- [ ] https://github.com/airqo-platform/AirQo-api/pull/2182
- [ ] https://github.com/airqo-platform/AirQo-api/pull/2184
- [ ] https://github.com/airqo-platform/AirQo-api/pull/2183

### DIFF
--- a/src/device-registry/utils/create-event.js
+++ b/src/device-registry/utils/create-event.js
@@ -24,6 +24,7 @@ const {
 } = require("./date");
 const { Parser } = require("json2csv");
 const httpStatus = require("http-status");
+const translateUtil = require("./translate");
 
 const listDevices = async (request) => {
   try {
@@ -538,6 +539,7 @@ const createEvent = {
       let limit = parseInt(query.limit) || 0;
       let skip = parseInt(query.skip);
       let filter = {};
+      const language = request.query.language;
 
       const responseFromFilter = generateFilter.events(request);
       if (responseFromFilter.success === true) {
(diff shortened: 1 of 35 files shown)


src/device-registry/utils/create-event.js
```diff
@@ -589,6 +591,16 @@ const createEvent = {
                 filter,
                 page,
               });
+              if (language !== undefined && constants.ENVIRONMENT === "STAGING ENVIRONMENT") {
+                let data = responseFromListEvents.data[0].data;
+                for (const event of data) {
```
Baalmart, line 596: Hi @BenjaminSsempala , this logic is still confusing me a little bit :)......
Especially the part of manipulating a const variable....and the logical flow of things is bit confusing!
Can we do this translation from somewhere else within the create-health-tips use case and minimise changes to this particular util function?

src/device-registry/utils/create-health-tips.js
```diff
@@ -24,12 +25,19 @@ const createHealthTips = {
...
+      const language = request.query.language;
+      let translatedHealthTips;
 
-      const responseFromListHealthTips = await HealthTipModel(tenant).list({
+      let responseFromListHealthTips = await HealthTipModel(tenant).list({
         filter,
         limit,
         skip,
       });
+      if (language !== undefined) {
```
Baalmart, line 36: Now, this logic is easier to understand than the one inside the create-events util. :) @BenjaminSsempala

src/device-registry/utils/create-event.js
BenjaminSsempala: @Baalmart The translation is also in the create-health-tips, however The tips in the mobile app come along with the measurements using the `/events` endpoint.

PR #2196, opened 2023-09-21
[META-DATA] JWT to token authentication

OchiengPaul442:
**_WHAT DOES THIS PR DO?_**
This Pull Request implements changes in the authentication method, transitioning from the use of JSON Web Tokens (JWT) to Token-based authentication.

No unit tests have been carried out on this PR.

N/A

### DIFF
--- a/src/meta-data/airqo_api.py
+++ b/src/meta-data/airqo_api.py
@@ -1,83 +1,41 @@
 import json
 import traceback
-
 import requests
-
 from config import Config
 
-
 class AirQoApi:
-    @staticmethod
-    def remove_suffix(string: str, suffix):
-        if string.endswith(suffix):
-            return string[: -len(suffix)]
-        else:
-            return string[:]
-
     def __init__(self):
-        self.AIRQO_BASE_URL = AirQoApi.remove_suffix(
-            Config.AIRQO_BASE_URL, "/")
+        self.AIRQO_BASE_URL = Config.AIRQO_BASE_URL.rstrip("/")
         self.API_AUTH_TOKEN = Config.API_AUTH_TOKEN
 
     def update_site_meta_data(self, site_details: dict):
-
         params = {"tenant": "airqo", "id": site_details.pop("id")}
-
-        for key, value in site_details.copy().items():
-            if value is None:
-                site_details.pop(key)
-
-        self.__request(
-            endpoint="devices/sites", params=params, body=site_details, method="put"
-        )
+        site_details = {k: v for k, v in site_details.items() if v is not None}
+        self.__request("devices/sites", params=params, body=site_details, method="PUT")
 
     def __request(self, endpoint, params=None, body=None, method=None):
-
         try:
             headers = {"Content-Type": "application/json"}
-            if method == "put":
-                api_request = requests.put(
-                    "%s/%s" % (self.AIRQO_BASE_URL, endpoint),
-                    params={
-                        **params, 'token': self.API_AUTH_TOKEN} if params else {'token': self.API_AUTH_TOKEN},
-                    headers=headers,
-                    data=json.dumps(body),
-                    verify=False,
-                )
-            elif method == "post":
-                api_request = requests.post(
-                    "%s/%s" % (self.AIRQO_BASE_URL, endpoint),
-                    params={
-                        **params, 'token': self.API_AUTH_TOKEN} if params else {'token': self.API_AUTH_TOKEN},
-                    headers=headers,
-                    data=json.dumps(body),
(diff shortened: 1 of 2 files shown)


Baalmart (commented):
Thanks @OchiengPaul442 , 

- Do we need to also update the SECRETS based on the new ENV I am seeing in the PR?
- Have we also reviewed the expiry date of the new token. 

Thanks!

OchiengPaul442:
Thank @Baalmart for your feedback,
Yes, We will need to include the new variable created in the secrets.
As for the token, I created one called MetaData that's for staging only and expires on April 19, 2024, so will need to generate another for production as well.

thanks.

Baalmart:
> Thanks @Baalmart for your feedback, Yes, We will need to include the new variable created in the secrets. As for the token, I created one called MetaData that's for staging only and expires on April 19, 2024, so will need to generate another for production as well.
> 
> thanks.

thanks @OchiengPaul442 ....please do so and leave them under my Slack DM so that I can update the corresponding SECRETS for BOTH production and staging ahead of any merge duty.

OchiengPaul442:
Okay

src/meta-data/airqo_api.py
```diff
@@ -33,21 +34,21 @@ def update_site_meta_data(self, site_details: dict):
     def __request(self, endpoint, params=None, body=None, method=None):
```
Baalmart, line 35: thanks @OchiengPaul442 :

Your attempt at refactoring the __request function to include the token as a query parameter looks good. It effectively adds the token parameter to the params dictionary for both PUT and POST requests, or creates a new params dictionary with only the token if params is None. This approach ensures that the token is included in the URL for authentication.

Here are a few suggestions for further improvement:

1. Error Handling: You're using a handle_api_error function, which is a good practice. However, it's essential to ensure that this function is defined and handles errors appropriately. You might consider passing an instance of your class to the __request method so that it can access the handle_api_error method within the class.

2. Code Reusability: If you have multiple methods that make API requests, you can further improve code reusability by centralizing common functionality like error handling and token handling.

3. Request URL: Constructing the URL using string concatenation (e.g., "%s/%s" % (self.AIRQO_BASE_URL, endpoint)) is fine, but you could use Python's f-strings (formatted strings) for improved readability. For example: f"{self.AIRQO_BASE_URL}/{endpoint}".

Here's a slightly modified version of your code with these suggestions in mind:

```
def __request(self, endpoint, params=None, body=None, method=None):
    try:
        headers = {"Content-Type": "application/json"}
        url = f"{self.AIRQO_BASE_URL}/{endpoint}"
        
        if params is None:
            params = {}
        params['token'] = self.API_AUTH_TOKEN

        if method == "put":
            api_request = requests.put(
                url,
                params=params,
                headers=headers,
                data=json.dumps(body),
                verify=False,
            )
        elif method == "post":
            api_request = requests.post(
                url,
                params=params,
                headers=headers,
                data=json.dumps(body),
                verify=False,
            )
        else:
            self.handle_api_error("Invalid")
            return None

        print(f"Request Url : {api_request.request.url}")
        print(f"Request Body : {api_request.request.body}")

        if api_request.status_code == 200 or api_request.status_code == 201:
            print(f"Request Response : {api_request.json()}")
        else:
            self.handle_api_error(api_request)
            return None
    except Exception as ex:
        print(ex)
        traceback.print_exc()
        return None

```

This code above above centralises token handling and URL construction, making it more concise and maintaining good readability.

Please review suggestions and share concerns/suggestions/thoughts accordingly.

Baalmart (commented):
Thanks @OchiengPaul442 for this. I have just a few comments for your review.

src/meta-data/airqo_api.py
OchiengPaul442: Thanks for the suggestions @Baalmart

OchiengPaul442: In the new commit, I have taken into account the new optimized code presented and also refactored a section to utilize the request function from the request library to reduce the code written and also make it more flexible for other HTTP requests.

```
if method == "put":
            api_request = requests.put(
                url,
                params=params,
                headers=headers,
                data=json.dumps(body),
                verify=False,
            )
        elif method == "post":
            api_request = requests.post(
                url,
                params=params,
                headers=headers,
                data=json.dumps(body),
                verify=False,
            )
        else:
            self.handle_api_error("Invalid")
            return None
```

Refactored to

`api_request = requests.request(method, url, params=params, headers=headers, data=json.dumps(body), verify=False)`

Baalmart (approved):

PR #2211, opened 2023-09-25
Fix MongoDB Error from cluster sharding setup - forecast job

Mnoble-19:
**_WHAT DOES THIS PR DO?_**
- Fixes error: 
```Failed to target upsert by query :: could not extract exact shard key, full error: {'index': 0, 'code': 61, 'errmsg': 'Failed to target upsert by query :: could not extract exact shard key'}```

### DIFF
--- a/src/airflow/airqo_etl_utils/ml_utils.py
+++ b/src/airflow/airqo_etl_utils/ml_utils.py
@@ -661,12 +661,7 @@ class ForecastUtils:
 
         for doc in forecast_results:
             try:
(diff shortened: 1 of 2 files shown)


Baalmart (approved):
Thanks @Mnoble-19 .....
Is this data for Forecasts also accessible via Big Query?

Mnoble-19:
> Thanks @Mnoble-19 ..... Is this data for Forecasts also accessible via Big Query?
Good morning @Baalmart , yes these forecasts are saved in BQ.

PR #2293, opened 2023-10-18
final routes for the verification emails

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [ ] final routes for the verification emails

### DIFF
--- a/src/auth-service/utils/email.templates.js
+++ b/src/auth-service/utils/email.templates.js
@@ -59,9 +59,9 @@ module.exports = {
     token,
     category,
   } = {}) => {
-    let url = `${constants.ANALYTICS_BASE_URL}/account/creation/step4/${user_id}/${token}/${category}`;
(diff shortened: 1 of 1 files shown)


src/auth-service/utils/email.templates.js
```diff
@@ -59,9 +59,9 @@ module.exports = {
     token,
     category,
   } = {}) => {
-    let url = `${constants.ANALYTICS_BASE_URL}/account/creation/step4/${user_id}/${token}/${category}`;
+    let url = `${constants.ANALYTICS_BASE_URL}/account/creation/individual/interest/${user_id}/${token}/${category}`;
```
Baalmart, line 62: Hi @FaithDaka , I have temporarily left the category path at the end of the URL since I am still trying to remember why it was put there in our initial work. Perhaps it is to be used by the endpoint you will be accessing?

But no worries since it wont affect any of the ongoing UI work, just ignore it for now. :)

PR #2301, opened 2023-10-19
Setup Mlflow on Kubernetes

Mnoble-19:
**_WHAT DOES THIS PR DO?_**
* Sets up AirQo Mlflow server on k8s (staging cluster)


k8s/mlflow/files/dbchecker.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
+
+if [[ -z "${PGHOST}" && -z "${MYSQL_HOST}" ]]; then
```
Baalmart, line 3: Thanks @Mnoble-19 
I would love to know more about this MYSQL setup. More like detailed notes about its future plans and all that. cc: @BenjaminSsempala

k8s/mlflow/templates/NOTES.txt
```diff
@@ -0,0 +1,22 @@
+1. Get the application URL by running these commands:
+{{/*{{- if false }}*/}}
```
Baalmart, line 2: Hi @Mnoble-19 , what is the significance of this file? Part of the config files? Or just to temporarily store some information?

Baalmart (commented):
thanks @Mnoble-19 , I have just left some few comments for your review.

k8s/mlflow/Chart.yaml
123MwanjeMike, line 1: I notice you have checked out [this chart](https://artifacthub.io/packages/helm/community-charts/mlflow) on ArtifactHub. It's actually possible to use that helm chart entirely without having to create a new chart. You can use [these parameters](https://artifacthub.io/packages/helm/community-charts/mlflow#values) in your values file.

k8s/mlflow/files/dbchecker.sh
Mnoble-19: > Thanks @Mnoble-19 I would love to know more about this MYSQL setup. More like detailed notes about its future plans and all that. cc: @BenjaminSsempala

Hello @Baalmart , no this is nothing to worry about. 
The bash script came with the helm chart and is for checking if one's  db is available, it's setup to check for either mysql or postgresql or both

in our scenario, we're using postgresql which is already setup on our VMs so no worries on mysql

k8s/mlflow/templates/NOTES.txt
Mnoble-19: Once the application is deployed, information about the node port and ip to access the application is printed in the terminal, using the contents of this file as a template, thus me committing it as well

k8s/mlflow/Chart.yaml
Mnoble-19: Hi @123MwanjeMike , so I've cleaned up the chart script with AirQo details only, and removed the default information. 
Hope that's okay

Psalmz777 (approved):

PR #2312, opened 2023-10-21
Rename airflow microservice to workflows

Mnoble-19:
**_WHAT DOES THIS PR DO?_**
* Renames the airflow microservice and its various references to `workflows`


.github/workflows/deploy-apis-to-production.yml
```diff
@@ -424,29 +424,29 @@ jobs:
         uses: docker/build-push-action@v4.1.1
         with:
           push: true
-          context: src/airflow
+          context: src/workflows
           target: deployment
-          tags: ${{ env.REGISTRY_URL }}/${{ env.PROJECT_ID }}/airqo-apache-airflow:${{ needs.image-tag.outputs.build_id }},${{ env.REGISTRY_URL }}/${{ env.PROJECT_ID }}/airqo-apache-airflow:latest
```
Mnoble-19, line 429: NB: Could cause some issues with deployment.

k8s/workflows/.helmignore
Mnoble-19, line 1: Hi @Baalmart , so for this part of k8s, I'm leaving it as it is for now, as the renaming, would also require removing the current deployment and redeploying afresh with the new names

CC: @Kisakye-e , @Psalmz777 , @BenjaminSsempala

Mnoble-19:
@Baalmart  I'm marking this as ready for review, but let's not merge so fast,

.github/workflows/deploy-apis-to-production.yml
Baalmart: Thanks @Mnoble-19 for the clarifications. Appears we might need to have bi-weekly DevOps workshops where we can collaboratively work on these things. ADDITIONALLY, we might want to work on them and test them over night times or weekends as a safety measure. cc: @BenjaminSsempala

Baalmart (commented):
thanks @Mnoble-19 ....looking forward to a DevOps workshop for us to review these changes. I have also left a couple of strategy approaches as review comments.

Baalmart:
Hi @Mnoble-19 , I notice that we have some merge conflicts which need to be resolved

PR #2357, opened 2023-10-26
Fix for predict caching mechanism

Mnoble-19:
**_WHAT DOES THIS PR DO?_**
*   Fixes issue of predict endpoints returning unexpected data by adding new variables to the `cache keys`. This will eliminate the issue of old cache values being returned, as is happening on the `production` heatmap


Baalmart (approved):
thanks @Mnoble-19

PR #2408, opened 2023-11-01
update the cronjob for updating the isActive status of users

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [ ] update the cronjob for updating the isActive status of users

### DIFF
--- a/src/auth-service/bin/cronJob.js
+++ b/src/auth-service/bin/cronJob.js
@@ -16,9 +16,16 @@ cron.schedule("0 0 * * *", async () => {
     while (true) {
       const users = await UserModel("airqo")
         .find({
-          lastLogin: {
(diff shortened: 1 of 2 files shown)


src/auth-service/bin/cronJob.js
```diff
@@ -16,9 +16,16 @@ cron.schedule("0 0 * * *", async () => {
...
-            $lt: new Date(Date.now() - inactiveThreshold),
-          },
+          $or: [
+            {
+              lastLogin: {
+                $lt: new Date(Date.now() - inactiveThreshold),
+              },
+            },
+            {
+              lastLogin: null,
```
Baalmart, line 26: include situation where the user has not yet logged into the system, just update their active status to false too

PR #2469, opened 2023-11-08
[auth service] user preferences and onboarding checklist

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [x] implements 2 new use cases of preferences and onboarding checklist

```
cd src/auth-service
npm install
npm run dev-mac
```

PREFERENCES

- [x] Create Preference
- [x] UPSERT Preference
- [x] Update Preference
- [x] List Preferences
- [x] List Preference Information
- [x] Delete Preference

CHECKLIST

- [x] Create Checklist
- [x] UPSERT Checklist
- [x] Update Checklist
- [x] Delete Checklist
- [x] List Checklists
- [x] List Checklist Information

USERS

- [ ] List User Information

### DIFF
--- a/src/auth-service/models/Preference.js
+++ b/src/auth-service/models/Preference.js
@@ -29,47 +29,47 @@ const PreferenceSchema = new mongoose.Schema(
       type: String,
       trim: true,
       required: [true, "pollutant is required!"],
-      preference: "pm2_5",
+      default: "pm2_5",
     },
     frequency: {
       type: String,
       required: [true, "frequency is required!"],
-      preference: "hourly",
+      default: "hourly",
     },
     startDate: {
       type: Date,
       required: [true, "startDate is required!"],
-      preference: addWeeksToProvideDateTime(currentDate, -2),
+      default: addWeeksToProvideDateTime(currentDate, -2),
     },
     endDate: {
       type: Date,
       required: [true, "endDate is required!"],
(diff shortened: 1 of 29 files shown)


src/auth-service/models/Preference.js
```diff
@@ -0,0 +1,341 @@
...
+      preference: mongoose.Types.ObjectId(constants.DEFAULT_GROUP),
+    },
+    user_id: {
+      type: ObjectId,
+      required: [true, "user_id is required"],
+      ref: "user",
+      unique: true,
+    },
+    site_ids: [
+      {
```
FaithDaka, line 95: @Baalmart Can we have an attribute for `selected_sites` instead of `site_ids`? So that all the data of the selected sites is stored at once, and we reduce the overhead involved in getting the `site_ids` then filtering and matching using data from the `sites` endpoints to get the sites data?

Baalmart: thanks @FaithDaka , just transitioning towards this.
Will reach out asap

Baalmart: @FaithDaka , 
our architecture is very separated in nature and the details of each Site ID are stored in a completely separate application.

So what you have suggested sounds good but it is technically impossible at the moment. 
What I promise to do is to continue to review and make enhancements as we move along

PR #2479, opened 2023-11-10
fixing the v2 of email verification

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [x] fixing the v2 of email verification

```
cd src/auth-service
npm install
npm run dev-mac
```

- [x] verify email: `{{baseUrl}}/api/v2/users/verify/{user_id}/{token}`

### DIFF
--- a/src/auth-service/routes/v2/users.js
+++ b/src/auth-service/routes/v2/users.js
@@ -285,7 +285,6 @@ router.post(
   createUserController.signUpWithFirebase
 );
 
-
 router.post(
   "/syncAnalyticsAndMobile",
   oneOf([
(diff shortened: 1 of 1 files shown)


src/auth-service/routes/v2/users.js
```diff
@@ -403,7 +394,7 @@ router.get(
 );
 
 router.get(
-  "/verify/:user_id/:token/:category",
```
Baalmart, line 406: @FaithDaka , just removed the "category" path which  was recently introduced for reasons I do not clearly remember. This is what made the endpoint "not available" as you earlier shared via Slack DM. cc: @BenjaminSsempala

PR #2499, opened 2023-11-14
introducing new fields for preferences

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [ ] introducing new fields for preferences

### DIFF
--- a/src/auth-service/models/Preference.js
+++ b/src/auth-service/models/Preference.js
@@ -47,6 +47,38 @@ const siteSchema = new mongoose.Schema(
   { _id: false }
 );
 
+const gridSchema = new mongoose.Schema(
+  {
+    _id: { type: ObjectId },
+    name: { type: String },
+  },
+  { _id: false }
+);
+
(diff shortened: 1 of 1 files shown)


src/auth-service/models/Preference.js
```diff
@@ -152,6 +184,11 @@ const PreferenceSchema = new mongoose.Schema(
       },
     ],
     selected_sites: [{ type: siteSchema }],
+    selected_grids: [{ type: gridSchema }],
+    selected_devices: [{ type: deviceSchema }],
+    selected_cohorts: [{ type: cohortSchema }],
+    selected_airqlouds: [{ type: airqloudSchema }],
+
```
Baalmart, line 191: Hi @FaithDaka , I have just introduced new array fields which take in objects for:

1. grids
2. devices
3. cohorts
4. airqlouds

Each of these objects just expects 2 things for now:

1. _id
2. name

cc: @Codebmk / @OchiengPaul442

PR #2509, opened 2023-11-16
reverse implementation for showing only public devices

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [ ] reverse implementation for showing only public devices -- requested by Joel due to some "new learnings". :)

### DIFF
--- a/src/device-registry/models/Event.js
+++ b/src/device-registry/models/Event.js
@@ -1337,7 +1337,7 @@ eventSchema.statics = {
 
       const startTime = filter["values.time"]["$gte"];
       const endTime = filter["values.time"]["$lte"];
-      const visibilityFilter = true;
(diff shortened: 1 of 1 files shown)


src/device-registry/models/Event.js
```diff
@@ -1551,7 +1551,7 @@ eventSchema.statics = {
             foreignField: "_id",
             as: "device_details",
           })
-          .match({ "device_details.visibility": visibilityFilter })
+          // .match({ "device_details.visibility": visibilityFilter })
```
Baalmart, line 1554: Hi @BenjaminSsempala , please take note of this commented out code in case such a request comes back in the near future from the hardware team. At the moment, they seem to be undecided hence the reason why we are reversing a change they initially asked us to make. :) cc: @Codebmk

PR #2703, opened 2023-12-28
do not send emails to the dummy mobile app unit test account

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [x] do not send emails to the dummy mobile app test account which is automated-tests@airqo.net. This particular account was spamming our email inbox for data@airqo.net with lots of undelivered notice emails since the email does not actually exist.

```
cd auth-service
npm install
npm run dev-mac
```

- [x] delete acccount
- [x] verify email

### DIFF
--- a/src/auth-service/utils/mailer.js
+++ b/src/auth-service/utils/mailer.js
@@ -281,6 +281,15 @@ const mailer = {
         attachments,
       };
 
+      if (email === "automated-tests@airqo.net") {
+        return {
+          success: true,
+          message: "email successfully sent",
+          data: [],
+          status: httpStatus.OK,
+        };
+      }
+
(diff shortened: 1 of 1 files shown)


src/auth-service/utils/mailer.js
```diff
@@ -350,6 +359,15 @@ const mailer = {
         };
       }
 
+      if (email === "automated-tests@airqo.net") {
```
Baalmart, line 362: @BenjaminSsempala , in each of our mailing functions, add an exception for this particular email address so that we always just return a success whenever it is encountered. 

In the near future of Q1 2024, we could also just enhance this and have it as an array of emails picked from a database so that we can future proof this part of adding exceptions for email sending.

PR #2756, opened 2024-01-05
Temporarily disable kafka in workflows

Mnoble-19:
**_WHAT DOES THIS PR DO?_**
* This PR temporarily disables Kafka in `workflows` for production and only limiting it to `staging`


Baalmart (approved):
@Mnoble-19

PR #2842, opened 2024-01-10
update mlflow port numbers

Mnoble-19:
**_WHAT DOES THIS PR DO?_**
- This changes the port number for the mlflow deployment as the previously chosen one is already used by `predict`

### DIFF
--- a/k8s/mlflow/values.yaml
+++ b/k8s/mlflow/values.yaml
@@ -7,8 +7,7 @@ nameOverride: ""
 fullnameOverride: ""
 service:
   type: NodePort
-  port: 5000
-  name: http
+  port: 5024
 
 backendStore:
   databaseConnectionCheck: true
@@ -24,11 +23,11 @@ artifactRoot:
 
 resources:
   limits:
-    cpu: 500m
-    memory: 1000Mi
+    cpu: 1000m
+    memory: 2000Mi
   requests:
-    cpu: 350m
-    memory: 700Mi
+    cpu: 800m
+    memory: 1500Mi
 
 extraVolumes:
   - name: mlflow-pv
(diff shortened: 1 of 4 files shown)


k8s/mlflow/values.yaml
```diff
@@ -24,11 +23,11 @@ artifactRoot:
 
 resources:
   limits:
-    cpu: 500m
-    memory: 1000Mi
+    cpu: 1000m
```
Baalmart, line 26: Thanks @Mnoble-19 for this.

Just a small observation here.....
We need to thoroughly review all resource increases accordingly. We need to avoid scenarios where users cannot do simple and extremely important actions like User Login (`auth-service`) because of `ml-flow`.

Looking forward to a details discussion around this topic and how we can continue to thoroughly separate concerns. cc: @Psalmz777 / @BenjaminSsempala

Mnoble-19: alright @Baalmart , I've decreased the resources allocated slightly. note that i actually started with small ones and gradually incremented before raising the PR on dev cluster,  also the app is only deployed to staging cluster ( no need for a 2nd prod version) so the above issues if they are to happen will only be limited to staging

Baalmart: Hi @Mnoble-19 .... thanks for the clarifications, much appreciated. :)

please note that staging is also a "production" for some of our stakeholders like the frontend devs. That is the main reason why I have been pushing for thee utilisation of the dev cluster.

Could we first go hard on mlflow (most of our ML solutions) via the dev cluster where we are free to do whatever we want?

How do we properly separate concerns here, could we perhaps repurpose the dev-cluster to be an ML cluster so that we start looking at it as a "production" of sorts? I look forward to more feedback and thoughts on this --- I really do not want to be fixing avoidable problems in the near future -- so as to allow to focus on new issues and also get time to learn new things as we continue to grow, professionally.

Baalmart (approved):

PR #2915, opened 2024-01-18
Resolves issue in failing weatherData dag

Mnoble-19:
**_WHAT DOES THIS PR DO?_**
- Resolves the issue of the failing weather data DAG. Adds import that may have been removed when renaming microservice


Baalmart (approved):
Thanks @Mnoble-19

PR #2960, opened 2024-01-24
 move to production

Baalmart:
- [x] https://github.com/airqo-platform/AirQo-api/pull/2959


Baalmart:
Hi @Psalmz777 / @BenjaminSsempala ....

- I have just updated the NGINX configurations. 
- I anticipate some instability within the next 4 hours or so.
- In case of any unusual instability, please just `restart `all the **production cluster** nodes (1,2 and 3).

PR #3038, opened 2024-02-15
[work in progress] move to production

Baalmart:
- [ ] https://github.com/airqo-platform/AirQo-api/pull/3037


Baalmart:
Hi @OchiengPaul442 , we shall need to create a new Page inside Netmanager for listing Client Registry where we manage clients accordingly. Management of clients will mainly include one thing: "Approving (activating) Clients using the Activate Client endpoint under [POST] `Authentication>>Clients>>Activate Client` on Postman."

Issue #629, opened 2021-10-11
track the battery status and make maintenance recommendations


Baalmart:
**Is your feature request related to a problem? Please describe.**
We need to be able to track the battery status and make recommendations on preventative maintenance. The rate of discharge can be tracked and display feedback to the users.

**Describe the solution you'd like**
We need to be able to track the battery status and make recommendations on preventative maintenance. The rate of discharge can be tracked and display feedback to the users.

**Describe alternatives you've considered**

- Create a notification systems for maintenance based on the discharge rate of a battery
- Showcase the discharge rate details in the device battery graphs

**Additional context**
This was picked from a hardware department member.

PR #3115, opened 2024-03-26
battery

marvinbanda:
battery analysis tool

issue  #[629  ](https://github.com/airqo-platform/AirQo-api/issues/629)

### DIFF
--- a/src/battery.html
+++ b/src/battery.html
@@ -1,0 +1,245 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


Baalmart (changes requested):
Thanks @marvinbanda for this PR.

A few notes for your consideration...

1. Our system is split into backend and frontend -- which makes it a 2 tier architecture, this particular repository is for our backend, so we do not expect frontend code here.
2. You need to divide your solution into backend (API) and frontend accordingly. This approach ensures re-usability of your solution in many other areas.

PR #3122, opened 2024-04-04
Setting up Virtual server Route for reports

BenjaminSsempala:
**_WHAT DOES THIS PR DO?_**

- [x] Setting up Virtual server Route for reports


Baalmart (commented):
thanks @BenjaminSsempala

PR #3219, opened 2024-06-07
Updated DAG definitions with correct parameters and schedule intervals

AnthonyByansi:
**_WHAT DOES THIS PR DO?_**

- [X] Correctly specified with `dag_id` and `schedule_interval` instead of schedule
- [X] start_date: Adds to ensure the DAGs start running from a specific date


Baalmart (approved):
thanks @AnthonyByansi

PR #3372, opened 2024-09-03
Fix Spatial Push to Production

BenjaminSsempala:
**_WHAT DOES THIS PR DO?_**

- [ ] Correct file paths for spatial action

**_HOW DO I TEST OUT THIS PR?_**

- [ ] Spatial: [Link to Spatial](https://github.com/airqo-platform/AirQo-api/tree/staging/src/spatial)

## Summary by CodeRabbit

- **Chores**
	- Updated the deployment configuration for production APIs to reflect changes in file paths.

PR #3373, opened 2024-09-03
move to production

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [x] https://github.com/airqo-platform/AirQo-api/pull/3372


.github/workflows/deploy-apis-to-production.yml
```diff
@@ -1628,7 +1628,7 @@ jobs:
         with:
           action: fjogeleit/yaml-update-action@main # Action to retry
           with: |
-            valueFile: "k8s/spatial-service/values-prod.yaml"
```
Baalmart, line 1631: this is a good catch, thanks @BenjaminSsempala

PR #3394, opened 2024-09-08
refactor Events Model & Util

Baalmart:
**_WHAT DOES THIS PR DO?_**

- [x] ensure that site details are in cohort ID mesurements
- [ ] ensure that BOTH device and site details are present in measurements when querying using cohorts
- [ ] nearest site endpoints should only return ACTIVE sites.
- [ ] Filter device measurements by the uptime threshold

To test locally, use` ?startTime=2022-12-11T22:28:11.245Z` or any other earlier date where you have measurements on your local machine.

- [x] Device Registry: [Link to Device Registry README](https://github.com/airqo-platform/AirQo-api/tree/staging/src/device-registry#readme)

- [x] Recent by Cohort ID: `{{baseUrl}}/api/v2/devices/measurements/cohorts/{COHORT_ID}/recent`

## Summary by CodeRabbit

- **New Features**
	- Updated event processing to categorize events using `site_id` instead of `device_id`, enhancing the accuracy of event identification.
  
- **Bug Fixes**
	- Resolved potential issues related to event categorization by ensuring the correct metadata is utilized in requests.


src/device-registry/controllers/create-event.js
```diff
@@ -2336,7 +2336,7 @@ const createEvent = {
         ? defaultTenant
         : req.query.tenant;
 
-      request.query.metadata = "device_id";
+      request.query.metadata = "site_id";
```
Baalmart, line 2339: Hi @Codebmk , please help me do some background work about which use cases utilise fetching measurements by Cohort ID.

This change ensures that we return site details when one fetches using  Cohort ID. Originally, we were returning Device Details since Cohort is very device specific.

On this note, we now need to find out the answers to the following questions:

1. Should we be returning both Site Details and Device Details?
2. Of the 2 (Sites and Devices), which one is more appropriate to be returned alongside measurements when one is fetching using Cohort ID?
3. Besides the researchers from Havard (Madagascar), who is currently using Cohort IDs? 
4. And how does this PR affect the people who are actively utilising this particular endpoint?

For more context, this PR is inspired by the issue you raised during the most recent car free day where you shared that getting measurements using Cohort ID does NOT return Site Details -- so this PR enables it. Just a bit hesitant in rushing the change hence my questions above.

I look forward to your timely feedback.

Codebmk: Hey @Baalmart , we don't need to return all the site details, but having the site name/search_name/description where the desired site details. And since this cohorts, it's meant to serve primarily device details. 

To answer your questions,
1. We don't need to return both site details and device details, normally device owners are more concerned about the device details
2. Device data is more appropriate when fetching from cohorts, but having few details about the site is also relevant esp site name/description
3. Not sure
4. As long we keep the device details as is, we should just be adding a few details and not removing anything

Baalmart: > Hey @Baalmart , we don't need to return all the site details, but having the site name/search_name/description where the desired site details. And since this cohorts, it's meant to serve primarily device details.
> 
> To answer your questions,
> 
> 1. We don't need to return both site details and device details, normally device owners are more concerned about the device details
> 2. Device data is more appropriate when fetching from cohorts, but having few details about the site is also relevant esp site name/description
> 3. Not sure
> 4. As long we keep the device details as is, we should just be adding a few details and not removing anything

Thanks @Codebmk , this is helpful!

PR #3909, opened 2024-11-25
move to production

Baalmart:
- [x] https://github.com/airqo-platform/AirQo-api/pull/3908

### DIFF
--- a/k8s/analytics/values-prod.yaml
+++ b/k8s/analytics/values-prod.yaml
@@ -8,7 +8,7 @@ images:
     celeryWorker: eu.gcr.io/airqo-250220/airqo-analytics-celery-worker
     reportJob: eu.gcr.io/airqo-250220/airqo-analytics-report-job
     devicesSummaryJob: eu.gcr.io/airqo-250220/airqo-analytics-devices-summary-job
-  tag: prod-eeb5fe06-1732531710
+  tag: prod-8e64a1e8-1732536341
 api:
   name: airqo-analytics-api
   label: analytics-api
--- a/k8s/auth-service/values-prod.yaml
+++ b/k8s/auth-service/values-prod.yaml
@@ -6,7 +6,7 @@ app:
 replicaCount: 3
 image:
   repository: eu.gcr.io/airqo-250220/airqo-auth-api
-  tag: prod-eeb5fe06-1732531710
+  tag: prod-8e64a1e8-1732536341
 nameOverride: ''
 fullnameOverride: ''
 podAnnotations: {}
--- a/k8s/device-registry/values-prod.yaml
+++ b/k8s/device-registry/values-prod.yaml
@@ -6,7 +6,7 @@ app:
 replicaCount: 3
 image:
   repository: eu.gcr.io/airqo-250220/airqo-device-registry-api
-  tag: prod-eeb5fe06-1732531710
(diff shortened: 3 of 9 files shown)


MikeKaiAI:
## Test Cleanup / Finalizer Analysis — Batch Lane Split Files

Investigated the three concerns about missing test-cleanup infrastructure after the split from the (now-deleted) `test_batch_lane_admin.py` into `test_batch_lane_case_door.py` and `test_batch_lane_org_defaults.py`.

### Finding 1 — `test_live_case_override_round_trip` has no cleanup finalizer

**Confirmed.** The new version in `test_batch_lane_case_door.py` (line 145) has this signature:

```python
def test_live_case_override_round_trip(env, http, evidence, super_token, lane_door_deployed):
```

There is no `request` parameter and no `addfinalizer` call. The test applies a hold at line 164 (`hold_hours: 1`) and then releases it at line 176 (`release: True`). If the test fails or errors **between** those two PATCH calls (e.g. the `assert hold.status_code == 200` passes but the release PATCH times out), the dev fixture case identified by `BATCH_LANE_TEST_CASE_ID` is left with an unreleased hold. No cleanup runs.

**Severity: Low (test-infrastructure only).** The test is already gated behind `env.is_prod` (skip on prod) and behind `BATCH_LANE_TEST_CASE_ID` being explicitly set (skip if unset), so the blast radius is a single developer's fixture case in dev. The hold is also only 1 hour, so it self-expires. But a `request.addfinalizer` that releases the hold on failure would be cleaner and avoid confusion during debugging.

### Finding 2 — No `_hold_applied` tracking variable

**Confirmed.** The old pattern used a `_hold_applied` flag set to `True` after the hold PATCH succeeds, and the finalizer conditionally released only when the flag was set. The new version has no such tracking. This is a direct consequence of Finding 1 — there's no finalizer, so there's nothing to condition on. If a finalizer were re-added, a tracking variable or try/finally pattern would again be appropriate to avoid releasing a hold that was never applied.

### Finding 3 — `test_valid_org_lane_round_trips` cleanup

**This one is actually fine — not a regression.** The new `test_batch_lane_org_defaults.py` replaced `request.addfinalizer(lambda: _restore_settings(...))` with the modern pytest **yield-fixture** pattern. The `org_settings_baseline` fixture (lines 112–129) does:

```python
yield original                                          # line 127
_restore_settings(http, env, super_token, original)     # line 129
```

This is functionally equivalent to `addfinalizer` — pytest runs the post-`yield` teardown code even if the test fails. The `test_valid_org_lane_round_trips` function (line 170) takes `org_settings_baseline` as a parameter, so the fixture's teardown is guaranteed to execute. **No issue here.**

### Summary

| Concern | Status | Risk |
|---|---|---|
| Missing hold-release finalizer in `test_live_case_override_round_trip` | **Confirmed — cleanup gap** | Low (dev-only, 1h self-expiring hold) |
| Missing `_hold_applied` tracking | **Confirmed — consequence of #1** | Low |
| Missing org-settings restore finalizer | **Not a regression** — replaced with equivalent yield-fixture | None |

All three are test-infrastructure concerns, not production correctness issues. The only actionable item is adding a cleanup finalizer (or try/finally) to `test_live_case_override_round_trip` in `test_batch_lane_case_door.py` to release any applied hold if the test fails mid-way.

PR #6417, opened 2026-04-16
Add gracefull shutdowns

NicholasTurner23:
Add container graceful shutdowns.


Baalmart (approved):
Hi @NicholasTurner23 

I've approved the PR and requested @Psalmz777  for a second review, as it may involve files Samantha updated yesterday to address system instability.

Best,
Martin

PR #6625, opened 2026-05-23
fix: update site hourly forecast logic to allow starting from current…

wabinyai:
… date and minimum record requirements

# 🚀 Pull Request

### What does this PR do?

- Updated site hourly forecasting to start forecast timestamps from the current run hour instead of each site’s latest historical measurement timestamp.

- Kept historical measurements as model input for lag and rolling features.

- Lowered the site hourly forecast minimum history requirement from 2 hourly records to 1 hourly record.

- Added/updated tests to verify sparse sites with one hourly record can still receive forecasts.

### Why is this change needed?

- Some sites had stale latest measurements, for example 17 May, causing forecast output timestamps to start from stale dates even when the DAG ran on 23 May.

- Forecasts should represent the current forecast horizon from today’s DAG run, while still using available historical data for prediction features.

- Allowing one historical hour improves coverage for newly deployed or sparse sites, while existing fallback logic fills missing lag/rolling features.

---

## 🔗 Related Issues
- [ ] Closes #
- [ ] Fixes #
- [ ] Related to #

---

## 🔄 Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🔧 Enhancement/improvement
- [ ] 📚 Documentation update
- [ ] ♻️ Refactor
- [ ] 🗑️ Removal/deprecation

---

## 🏗️ Affected Services
**Microservices changed:**

---

## 🧪 Testing
- [ ] Unit tests added/updated
- [ ] Manual testing completed
- [ ] All existing tests pass

**Test summary:**

---

## 💥 Breaking Changes
- [ ] **No breaking changes**
- [ ] **Has breaking changes** (describe below)

---

## 📝 Additional Notes

---

## ✅ Checklist
- [ ] Code follows project style guidelines
- [ ] Self-review completed
- [ ] Documentation updated (if needed)
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Hourly site forecasts can now start from an explicit forecast start timestamp to control forecasting boundaries.

* **Improvements**
  * Site eligibility relaxed from requiring 2+ historical hourly records to 1+ within the trimmed history window.
  * Forecast retention uses a configurable day-based cutoff rather than an hour-based horizon.

* **Chores**
  * Added SITE_HOURLY_FORECAST_RETENTION_CUT_OFF_DAY env var (default: 2 days).

* **Tests**
  * Updated and added tests to verify start-timestamp behavior and new retention pruning.

[Review Change Stack](https://app.coderabbit.ai/change-stack/airqo-platform/AirQo-api/pull/6625?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Baalmart (commented):
Thanks @wabinyai 

Please review the following feedback from the BOT

Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

Outside diff comments:
In `@src/workflows/airqo_etl_utils/ml_utils.py`:
- Around line 2423-2429: The new logic in the history filter allows sites with a
single record, but the helper
ForecastModelTrainer.fetch_site_hourly_prediction_data still defaults
min_hours=2 so callers will continue to drop those sites upstream; update the
fetch_site_hourly_prediction_data signature to default min_hours=1 (or
alternatively leave the default and update every call that must enforce the
stricter rule to explicitly pass min_hours=2), and ensure any callers that
relied on the old behavior are updated to pass min_hours=2 where needed so the
downstream history filtering (the block that builds site_counts and checks
history.empty) and the similar block at the later location behave consistently
end-to-end.

---

Nitpick comments:
In `@src/workflows/airqo_etl_utils/tests/test_site_hourly_forecast.py`:
- Around line 420-423: The pruning tests compute expected_cutoff from the live
clock and an assumed 2-day retention, which makes them flaky; instead patch the
time and retention config so expected_cutoff is deterministic: freeze or
monkeypatch pd.Timestamp.now (or use freezegun) to a fixed UTC datetime and, if
retentionDays comes from config, monkeypatch the config value used by
ForecastModelTrainer.save_site_hourly_forecasts_to_mongo to the expected 2 days
before calling the function, then compute expected_cutoff from that fixed now so
the assertions are stable.
- Around line 107-119: Add a per-site assertion that the first forecast
timestamp equals the provided run_timestamp to ensure anchoring uses the run
hour rather than last observation; after calling
ForecastModelTrainer.generate_site_hourly_forecasts (the variable forecasts)
assert that for every site_id the minimum timestamp equals
pd.Timestamp("2026-03-04T08:00:00Z") (or compare to the run_timestamp variable)
so each site's first forecast hour is the run hour.

NicholasTurner23 (approved):

